PR #3, opened 2020-08-12
Add validation to create and update

rmweir:
**Problem:**
Users need the option to upgrade nodegroup kubernetes versions the amazon way or to create new nodegroups and delete ones that are outdated.

**Solution:**
Nodegroup versions must now be maintained. Control plane cannot upgraded until all nodegroup versions match. New nodegroups' versions must match the control plane's version. Nodegroup versions can at most be 1 minor version less than the control plane's version.

**Issue:**
https://github.com/rancher/rancher/issues/27777

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/eks"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/blang/semver"
 	v13 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
 	v14 "github.com/rancher/eks-operator/pkg/generated/controllers/core/v1"
 	v12 "github.com/rancher/eks-operator/pkg/generated/controllers/eks.cattle.io/v1"
@@ -111,6 +112,10 @@ func (h *Handler) recordError(onChange func(key string, config *v13.EKSClusterCo
 			}
 			config = config.DeepCopy()
 			config.Status.FailureMessage = err.Error()
+			if config.Status.Phase == eksConfigActivePhase {
+				// can assume an update is failing
+				config.Status.Phase = eksConfigUpdatingPhase
+			}
 			var recordErr error
 			config, recordErr = h.eksCC.UpdateStatus(config)
 			if recordErr != nil {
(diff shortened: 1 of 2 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -246,6 +251,18 @@ func deleteNodeGroups(clusterName string, eksService *eks.EKS, nodeGroups []v13.
...
 func (h *Handler) checkAndUpdate(config *v13.EKSClusterConfig, eksService *eks.EKS, svc *cloudformation.CloudFormation) (*v13.EKSClusterConfig, error) {
+	if err := validateUpdate(config); err != nil {
+		// validation failed, will be considered a failing update until resolved
+		config = config.DeepCopy()
+		config.Status.Phase = eksConfigUpdatingPhase
+		var updateErr error
+		config, updateErr = h.eksCC.UpdateStatus(config)
+		if updateErr != nil {
+			return config, updateErr
+		}
```
aiyengar2, line 262: nit:
```suggestion
		if config, updateErr := h.eksCC.UpdateStatus(config); updateErr != nil {
			return config, updateErr
		}
```

controller/eks-cluster-config-handler.go
```diff
@@ -371,7 +388,39 @@ func createStack(svc *cloudformation.CloudFormation, name string, displayName st
...
+			errors = append(errors, fmt.Sprintf("improper version format for nodegroup [%s]: %s", ng.NodegroupName, aws.StringValue(ng.Version)))
+		}
+		if clusterVersion.EQ(*version) {
+			continue
+		}
+		if clusterVersion.Minor-version.Minor == 1 {
+			continue
+		}
+		errors = append(errors, fmt.Sprintf("versions for cluster [%s] and nodegroup [%s] not compatible: all nodegroup kubernetes version"+
+			"must be equal to or one minor version lower than cluster kubernetes version", config.Spec.KubernetesVersion, aws.StringValue(ng.Version)))
```
aiyengar2, line 411: nit:

```suggestion
		errors = append(errors, fmt.Sprintf("versions for cluster [%s] and nodegroup [%s] not compatible: all nodegroup kubernetes versions"+
			"must be equal to or one minor version lower than cluster kubernetes version", config.Spec.KubernetesVersion, aws.StringValue(ng.Version)))
```

controller/eks-cluster-config-handler.go
rmweir: I suppose I could still do this but I purposely did not because I want configs value to outlive this scope and usually use this syntax to contain.

controller/eks-cluster-config-handler.go
```diff
@@ -371,7 +388,39 @@ func createStack(svc *cloudformation.CloudFormation, name string, displayName st
...
+	// validate nodegroup versions
+	for _, ng := range config.Spec.NodeGroups {
+		version, err := semver.New(fmt.Sprintf("%s.0", aws.StringValue(ng.Version)))
+		if err != nil {
+			errors = append(errors, fmt.Sprintf("improper version format for nodegroup [%s]: %s", ng.NodegroupName, aws.StringValue(ng.Version)))
+		}
+		if clusterVersion.EQ(*version) {
+			continue
+		}
+		if clusterVersion.Minor-version.Minor == 1 {
```
cbron, line 407: wouldn't this be 1 or 0 ? If they are on same version would be fine ?

rmweir: that what the EQ above this is checking

PR #4, opened 2020-08-14
Always set local networking variables

rmweir:
**Problem:**
Local networking fields were only being copied from status if status was recently updated. If the object reentered the create
logic after the networking fields on status are updated, the local networking variables are left blank. This causes errors during cluster creation in EKS.

**Solution:**
Always copy networking fields from status.

**Issue:**
https://github.com/rancher/rancher/issues/28129

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -432,54 +432,14 @@ func (h *Handler) create(config *v13.EKSClusterConfig, sess *session.Session, ek
 
 	displayName := config.Spec.DisplayName
 
-	var subnetIds []*string
-	var securityGroups []*string
-	if len(config.Spec.Subnets) != 0 {
-		logrus.Infof("VPC info provided, skipping vpc/subnet/securitygroup creation")
-		config = config.DeepCopy()
-		// copy networking fields to status
-		config.Status.Subnets = config.Spec.Subnets
-		config.Status.SecurityGroups = config.Spec.SecurityGroups
-		config.Status.NetworkFieldsSource = "provided"
-		var err error
-		config, err = h.eksCC.UpdateStatus(config)
-		if err != nil {
-			return config, err
-		}
-
-		subnetIds = aws.StringSlice(config.Spec.Subnets)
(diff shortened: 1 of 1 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -495,6 +455,56 @@ func (h *Handler) create(config *v13.EKSClusterConfig, sess *session.Session, ek
...
+		// copy networking fields to status
+		config.Status.Subnets = config.Spec.Subnets
+		config.Status.SecurityGroups = config.Spec.SecurityGroups
+		config.Status.NetworkFieldsSource = "provided"
+		var err error
+		config, err = h.eksCC.UpdateStatus(config)
+		if err != nil {
+			return config, err
+		}
+	} else {
```
aiyengar2, line 476: Can we remove the else statement here? It seems like at the moment we plan to `return h.eksCC.UpdateStatus(config)` regardless of the output (either returning `config, err` where `err != nil` in line 474 or `config, nil` where `err == nil` in line 505).

controller/eks-cluster-config-handler.go
```diff
@@ -495,6 +455,56 @@ func (h *Handler) create(config *v13.EKSClusterConfig, sess *session.Session, ek
...
+		config.Status.SecurityGroups = strings.Split(securityGroupsString, ",")
+		config.Status.Subnets = strings.Split(subnetIdsString, ",")
+		config.Status.NetworkFieldsSource = "generated"
+		config, err = h.eksCC.UpdateStatus(config)
+		if err != nil {
+			return config, err
+		}
+	}
+
+	return config, nil
```
aiyengar2, line 505: Once we remove the else (as per my previous comment), we could just replace this with `return h.eksCC.UpdateStatus(config)`

rmweir: We cannot remove the else and move the updateStatus to the bottom because if the first condition is met the fields will be overwritten by the logic currently contained in the else. We can either return early with update in the first block or move the overlapping update from both to the bottom.

rmweir: Went with keeping the else and returning the overlapping update status at the bottom.

PR #5, opened 2020-08-16
Add ng update options

rmweir:
**Problem:**
Cannot update nodegroup fields: min size, max size, desired size, labels, tags.

**Solution:**
Detect changes between aforementioned fields and upstream and update fields that do not match.

**Issues:**
https://github.com/rancher/rancher/issues/28153
https://github.com/rancher/rancher/issues/28151
https://github.com/rancher/rancher/issues/28126

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -289,6 +289,7 @@ func (h *Handler) checkAndUpdate(config *v13.EKSClusterConfig, eksService *eks.E
 
 	// gather upstream node groups states
 	var nodeGroupStates []*eks.DescribeNodegroupOutput
+	nodegroupARNs := make(map[string]string)
 	for _, ngName := range ngs.Nodegroups {
 		ng, err := eksService.DescribeNodegroup(
 			&eks.DescribeNodegroupInput{
@@ -314,6 +315,7 @@ func (h *Handler) checkAndUpdate(config *v13.EKSClusterConfig, eksService *eks.E
 		}
 
 		nodeGroupStates = append(nodeGroupStates, ng)
+		nodegroupARNs[aws.StringValue(ngName)] = aws.StringValue(ng.Nodegroup.NodegroupArn)
 	}
 
 	upstreamSpec, clusterARN, err := h.buildUpstreamClusterState(config.Spec.DisplayName, clusterState, nodeGroupStates, eksService)
@@ -321,7 +323,7 @@ func (h *Handler) checkAndUpdate(config *v13.EKSClusterConfig, eksService *eks.E
 		return config, err
 	}
 
-	return h.updateUpstreamClusterState(upstreamSpec, config, clusterARN, eksService, svc)
+	return h.updateUpstreamClusterState(upstreamSpec, config, clusterARN, nodegroupARNs, eksService, svc)
 }
 
 func createStack(svc *cloudformation.CloudFormation, name string, displayName string,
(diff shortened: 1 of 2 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -883,26 +885,87 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *v13.EKSClusterConfigS
...
+
+		if sendUpdateNodegroupConfig {
+			attemptUpdatingNodegroups = true
+			_, err := eksService.UpdateNodegroupConfig(updateNodegroupConfig)
+			if err != nil {
+				return config, err
+			}
+			continue
+		}
+
```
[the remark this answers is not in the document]
rmweir: add comment on L890 and L980

controller/eks-cluster-config-handler.go
```diff
@@ -883,26 +885,87 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *v13.EKSClusterConfigS
...
+		var sendUpdateNodegroupConfig bool
+
+		if unlabels := getUpdateUntags(aws.StringValueMap(ng.Labels), aws.StringValueMap(upstreamNg.Labels)); unlabels != nil {
+			updateNodegroupConfig.Labels = &eks.UpdateLabelsPayload{
+				RemoveLabels: unlabels,
+			}
+			sendUpdateNodegroupConfig = true
+		}
+
+		if labels := getUpdateTags(aws.StringValueMap(ng.Labels), aws.StringValueMap(upstreamNg.Labels)); labels != nil {
```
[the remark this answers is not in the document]
rmweir: These functions can be used for any two maps. Changed names for getUpdateTags and getUpdateUntags to GetKeyValuesToUpdate and GetKeysToDelete respectively and moved to the util package.

utils/map.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func GetKeysToDelete(tags map[string]string, upstreamTags map[string]string) []*string {
+	if len(upstreamTags) == 0 {
+		return nil
+	}
+
+	var updateUntags []*string
+	for key, val := range upstreamTags {
+		if len(tags) == 0 || tags[key] != val {
```
mrajashree, line 36: if len(tags) is 0, do you think we can check this before the for loop and return nil?

controller/eks-cluster-config-handler.go
```diff
@@ -883,26 +885,101 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *v13.EKSClusterConfigS
...
-	for _, ng := range upstreamSpec.NodeGroups {
-		if aws.StringValue(ng.Version) == desiredNgVersions[ng.NodegroupName] {
+	var attemptUpdatingNodegroups bool
+	for _, upstreamNg := range upstreamSpec.NodeGroups {
+		// if continue is used after an update, it means that update
+		// must finish before others for that nodegroup can take place.
+		// Some updates such as minSize, maxSize, and desiredSize can
+		// happen together
+
+		ng := ngs[upstreamNg.NodegroupName]
```
mrajashree, line 895: do we need to check if this upstream node group exists in the map?

utils/map.go
rmweir: No because if len(tags) is 0 then we want to delete all of the keys.

controller/eks-cluster-config-handler.go
rmweir: No because above this is logic to create and then return if the nodegroup is not in the upstream map. This point will not be reached if that's the case.

utils/map.go
mrajashree: oh I misread this, makes sense

PR #8, opened 2020-08-26
Access update fix

rmweir:
**Problem:**
Update from public access to private access if not working. Public and private access updates need to be sent together. When they are sent one at a time the request may be denied due to having both public and private access disabled.

Other commits are currently under review in other PR. Cherry-picked so conflicts are easier to resolve later.

**Solution:**
Send public and private access updates together.

**Issue:**
https://github.com/rancher/rancher/issues/28115


brendarearden (approved):
looks good to me!

paynejacob (approved):

cbron (commented):
Only reviewed last commit but that looks fine, need to fix conflict.

PR #9, opened 2020-08-27
Add tag to subnets for elb compatibility

rmweir:
**Problem:**
Need networking tags for load balancer and elb compatibility in EKS.

**Solution:**
Add following tag:
kubernetes.io/role/elb: 1

**Notes:**
"kubernetes.io/cluster" tag is added by EKS. "kubernetes.io/role/internal-elb" does not apply because only public subnets are created for the user.

**Issue:**
https://github.com/rancher/rancher/issues/28332

### DIFF
--- a/templates/template.go
+++ b/templates/template.go
@@ -117,6 +117,8 @@ Resources:
       Tags:
       - Key: Name
         Value: !Sub "${AWS::StackName}-Subnet01"
+      - Key: kubernetes.io/role/elb
+        Value: 1
 
   Subnet02:
     Type: AWS::EC2::Subnet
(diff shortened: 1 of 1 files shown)


brendarearden:
is the user able to create private subnets later and if so, would they have to add the tags themselves?

rmweir:
@brendarearden you cannot change the subnets that are assigned to an EKS cluster. They can however user their own subnets to create the cluster. These would need to be tagged by the user. @cloudnautique just wants the one we generate to be properly tagged.

rmweir:
@cbron nope, that was an accident. That's meant for a different issue. Removed.

PR #11, opened 2020-09-09
Check length of instance types

rmweir:
**Problem:**
Prior, importing clusters with nodegroups that don't have instance
types on their state would cause a nil. Nodegroups will not have
instance types if they are using a launch template. This is because
the instance types are instead described on the launch template.

**Solution:**
Now, the instance type will only be copied if it exists on the
nodegroup state.

**Issue:**
https://github.com/rancher/rancher/issues/28801


brendarearden (approved):
looks good!

paynejacob (approved):

cbron (approved):

PR #13, opened 2020-09-17
Add ssh key to upstream state

rmweir:
**Problem:**
Ssh key is showing nil instead of proper value on spec. This is because the ssh key is not being included when building the upstream spec and then the upstream spec is being used to overwrite spec.

**Solution:**
Add ssh key when building upstream state.

**Issue:**
https://github.com/rancher/rancher/issues/28937


brendarearden (approved):
Looks good!

paynejacob (approved):

cbron (approved):

PR #14, opened 2020-09-17
Require non-nil version for nodegroups

rmweir:
**Problem:**
Nodegroup version should be provided on create because it is required in EKS.

**Solution:**
Add validation that returns an error if a nodegroup's version is nil on create.

**Issue:**
https://github.com/rancher/rancher/issues/27777


paynejacob (approved):

brendarearden (approved):
LGTM

cbron (approved):

PR #16, opened 2020-09-21
Add owner reference to secret

rmweir:
**Problem:**
Secrets created by eks-operator were not being cleaned up after corresponding EKSClusterConfig was deleted.

**Solution:**
Add EKSClusterConfig as owner reference

**Issue:**
https://github.com/rancher/rancher/issues/29040

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -16,7 +16,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/eks"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/blang/semver"
-	v13 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
+	eksv1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
 	v12 "github.com/rancher/eks-operator/pkg/generated/controllers/eks.cattle.io/v1"
 	"github.com/rancher/eks-operator/templates"
 	"github.com/rancher/eks-operator/utils"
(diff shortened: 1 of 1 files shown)


paynejacob (approved):

controller/eks-cluster-config-handler.go
```diff
@@ -1132,7 +1133,7 @@ func (h *Handler) importCluster(config *v13.EKSClusterConfig, eksService *eks.EK
 
 // createCASecret creates a secret containing ca and endpoint. These can be used to create a kubeconfig via
 // the go sdk
-func (h *Handler) createCASecret(name, namespace string, clusterState *eks.DescribeClusterOutput) error {
+func (h *Handler) createCASecret(name, namespace string, config *eksv1.EKSClusterConfig, clusterState *eks.DescribeClusterOutput) error {
```
aiyengar2, line 1136: Is the `name` and `namespace` provided to this function always `config.Name` or `config.Namespace`? If so, we should remove the `name` and `namespace` from this since it already takes in the `config`.

aiyengar2 (changes requested):

paynejacob (approved):

aiyengar2 (approved):

controller/eks-cluster-config-handler.go
aiyengar2: seems like this was updated in the latest commit

cbron (approved):

controller/eks-cluster-config-handler.go
rmweir: Yes good catch, fixed.

PR #18, opened 2020-10-21
Disallow creating cluster if one with same name exists

thedadams:
Before creating a new hosted EKS cluster, we first check the existing
clusters to ensure the name provided by the user is unique. If it is not
unique, then we wait for user to delete the cluster in Rancher or the
existing cluster in EKS.

If the cluster never advanced passed the "not created" status, then we
also should not delete any existing EKS clusters.

Issue:
https://github.com/rancher/rancher/issues/29103

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -147,6 +147,11 @@ func (h *Handler) OnEksConfigRemoved(key string, config *eksv1.EKSClusterConfig)
 		logrus.Infof("cluster [%s] is imported, will not delete EKS cluster", config.Name)
 		return config, nil
 	}
+	if config.Status.Phase == eksConfigNotCreatedPhase {
+		// The most likely context here is that the cluster already existed in EKS, so we shouldn't delete it
+		logrus.Warnf("cluster [%s] never advanced to creating status, will not delete EKS cluster", config.Name)
+		return config, nil
+	}
 
 	logrus.Infof("deleting cluster [%s]", config.Name)
 
@@ -430,7 +435,7 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
 }
 
 func (h *Handler) create(config *eksv1.EKSClusterConfig, sess *session.Session, eksService *eks.EKS) (*eksv1.EKSClusterConfig, error) {
-	if err := validateCreate(config); err != nil {
+	if err := h.validateCreate(config, eksService); err != nil {
 		return config, err
 	}
 
@@ -443,8 +448,8 @@ func (h *Handler) create(config *eksv1.EKSClusterConfig, sess *session.Session, 
 	svc := cloudformation.New(sess)
 
 	displayName := config.Spec.DisplayName
-
 	var err error
+
 	config, err = h.generateAndSetNetworking(svc, config)
 	if err != nil {
 		return config, err
@@ -517,7 +522,28 @@ func (h *Handler) create(config *eksv1.EKSClusterConfig, sess *session.Session, 
 	return h.eksCC.UpdateStatus(config)
 }
 
-func validateCreate(config *eksv1.EKSClusterConfig) error {
+func (h *Handler) validateCreate(config *eksv1.EKSClusterConfig, eksService *eks.EKS) error {
+	// Check for existing eksclusterconfigs with the same display name
+	eksConfigs, err := h.eksCC.List(config.Namespace, v15.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("cannot list eksclusterconfigs for display name check")
+	}
+	for _, c := range eksConfigs.Items {
+		if c.Spec.DisplayName == config.Spec.DisplayName && c.Name != config.Name {
+			return fmt.Errorf("cannot create cluster [%s] because an eksclusterconfig exists with the same name", config.Spec.DisplayName)
+		}
+	}
+	// Check for existing clusters in EKS with the same display name
+	listOutput, err := eksService.ListClusters(&eks.ListClustersInput{})
+	if err != nil {
+		return fmt.Errorf("error listing clusters: %v", err)
+	}
+	for _, cluster := range listOutput.Clusters {
+		if aws.StringValue(cluster) == config.Spec.DisplayName {
+			return fmt.Errorf("cannot create cluster [%s] because a cluster in EKS exists with the same name", config.Spec.DisplayName)
+		}
+	}
+
 	// validate nodegroup version
 	if !config.Spec.Imported {
 		cannotBeNilError := "field [%s] cannot be nil for non-import cluster [%s]"


rmweir (commented):
This looks good, although I think a change may be required in Rancher before this would actually show in the UI. Did you test that?

controller/eks-cluster-config-handler.go
```diff
@@ -444,7 +449,17 @@ func (h *Handler) create(config *eksv1.EKSClusterConfig, sess *session.Session,
...
 	displayName := config.Spec.DisplayName
 
-	var err error
+	listOutput, err := eksService.ListClusters(&eks.ListClustersInput{})
+	if err != nil {
+		return config, fmt.Errorf("error listing clusters: %v", err)
+	} else {
+		for _, cluster := range listOutput.Clusters {
+			if cluster != nil && *cluster == displayName {
+				return config, fmt.Errorf("cannot create cluster with same name as an existing cluster")
```
rmweir, line 458: you should mention which cluster this is referencing
```suggestion
				return config, fmt.Errorf("cannot create cluster [%s] because a cluster exists in EKS with that name")
```

rmweir (changes requested):

thedadams:
The error message displays in the UI without any changes to Rancher. That's why I didn't realize I should add the name in the error message because the UI has the name as well.

rmweir:
@thedadams oh you're right there is logic for that phase in the Rancher integration controller. The name is useful when someone is making sense out of the actual logs, especially when using the eks-operator outside of Rancher.

rmweir (approved):
LGTM

controller/eks-cluster-config-handler.go
```diff
@@ -147,6 +147,11 @@ func (h *Handler) OnEksConfigRemoved(key string, config *eksv1.EKSClusterConfig)
 		logrus.Infof("cluster [%s] is imported, will not delete EKS cluster", config.Name)
 		return config, nil
 	}
+	if config.Status.Phase == eksConfigNotCreatedPhase {
+		// The most likely context here is that the cluster already existed in EKS, so we shouldn't delete it
+		logrus.Infof("cluster [%s] never advanced to creating status, will not delete EKS cluster")
+		return config, nil
+	}
```
cbron, line 154: Assuming this would also work with imported. Wouldn't we also want to show this message to user in the UI ?

cbron (commented):
main part seems fine

controller/eks-cluster-config-handler.go
thedadams: Once a cluster is deleted, it is removed from the UI pretty quickly. I believe that this message would only flash on the UI for a moment unless something in the UI changed.

cbron: Makes sense, @rmweir maybe you have another idea. We could run into a scenario where someone deletes the cluster and it sticks around without them knowing, that costs money. If nothing else we should log. Maybe bump to warn ?

cbron (approved):
lgtm besides log error, but wait to merge until 2.5.2 goes out

controller/eks-cluster-config-handler.go
thedadams: Just a thought: the message that is displayed in the UI could be changed to: "cannot create cluster [%s] because a cluster in EKS exists with the same name, removing the cluster from Rancher will have no effect on the existing one"

rmweir: @cbron makes a good point though, maybe the cluster has the same name but we still brought up a vpc for it and in that case should we tear it down? would require some more code since it doesn't really differentiate if it exists or it had to create it right now. Easiest way would be to not remove anything but communicate that.

thedadams: I understand what you are saying: we don't want to waste any money. Right now, we check for an existing cluster before anything is created. Is it enough to maintain this going forward?

controller/eks-cluster-config-handler.go
```diff
@@ -444,7 +449,16 @@ func (h *Handler) create(config *eksv1.EKSClusterConfig, sess *session.Session,
...
+	listOutput, err := eksService.ListClusters(&eks.ListClustersInput{})
+	if err != nil {
+		return config, fmt.Errorf("error listing clusters: %v", err)
+	}
+	for _, cluster := range listOutput.Clusters {
+		if aws.StringValue(cluster) == displayName {
+			return config, fmt.Errorf("cannot create cluster [%s] because a cluster in EKS exists with the same name", displayName)
+		}
+	}
+
```
prachidamle, line 461: Don't we need to protect from multiple create calls that happen parallely? Its possible to use the same cluster name if the create happens almost at the same time. Any lock needed around this?

thedadams: @rmweir Is that something we need to worry about for this? I know there is some login in rancher to disallow creating a cluster with the same name as one that already exists in rancher, but not sure if that is concurrent-safe.

rmweir: There is a validator which seeks to prevent this. Since it uses the cache it is possible but unlikely this would happen unless someone was programmatically making clusters that reference the same EKS cluster. Something that could happen is an object being reenqueued due to an update below and going through this block again before being set to "active". However, I think this is also unlikely and not sure an UpdateStatus call causes a reenqueue in rancher; just something to keep an eye on.

prachidamle: We could just use a local mutex to lock the creation - this would save from concurrent calls made to same server instance - [For an HA setup, concurrent calls can still happen if they landup on different instances.]

rmweir: @prachidamle Idk how this would be more prevalent on an HA setup since only one instance is deployed and the deployment code is launch from the management context. I don't think we should be adding mutex's premptively since we don't have evidence this would occur.

prachidamle: Oh good if there will be only single instance in HA then a mutex would definitely work for parallel requests coming in. We should test this by making parallel create calls and using same clustername.

prachidamle: We have used such a mutex, in Rancher's cluster store earlier

thedadams: I'm not sure if `eks-operator` is the right place for the mutex because the request isn't going to `eks-operator`. The request goes to `rancher`, it creates the `eksoperatorconfig` resource, then `eks-operator` gets alerted to the existence of the resource.

I'm still learning about the `rancher` API, but perhaps the very mutex your linked to will handle the case you are referring to.

prachidamle: Ok, is it possible to test it out? If it's the Rancher API call, we can call it parallely to see if same name can get used by eks?

thedadams: Yes. I tested that the mutex in the linked code protects against multiple clusters resources from being created with the same display name.

controller/eks-cluster-config-handler.go
rmweir: I think that should prevent this.

rmweir: I think that should handle this, we may run into situation in the future where it doesn't get into "creating" for other reasons like an aws limit, so we should keep that in mind at least.

rmweir: nit: maybe we should leave a comment here explaining

rmweir (approved):
LGTM with one nit and anticipating that there may be situations where the operator does not clean up non-cluster resources.

controller/eks-cluster-config-handler.go
```diff
@@ -444,7 +449,16 @@ func (h *Handler) create(config *eksv1.EKSClusterConfig, sess *session.Session,
...
-	var err error
+	listOutput, err := eksService.ListClusters(&eks.ListClustersInput{})
+	if err != nil {
+		return config, fmt.Errorf("error listing clusters: %v", err)
+	}
+	for _, cluster := range listOutput.Clusters {
+		if aws.StringValue(cluster) == displayName {
+			return config, fmt.Errorf("cannot create cluster [%s] because a cluster in EKS exists with the same name", displayName)
+		}
+	}
```
rmweir, line 460: Can you throw a comment above here or put it in a function with a name describing what it's doing? I think at first glance someone could be confused as to how this and what is done in validateCreate is different.

rmweir (changes requested):

rmweir (approved):
LGTM

PR #20, opened 2020-10-28
Add CI script

rmweir:
**Problem:**
Sometimes code is committed without running go generate.

**Solution:**
Include CI that runs go generate and checks for diffs.

**Issue:**
https://github.com/rancher/rancher/issues/29294


thedadams (approved):

thedadams (approved):

scripts/ci
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+cd $(dirname $0)/..
+
+echo Running: go fmt
```
aiyengar2, line 5: nit: should we wrap these with quotes?
```suggestion
echo "Running: go fmt"
```

rmweir: Added single quotes.

cbron (approved):
Seems like a good start

PR #21, opened 2020-10-30
Remove creation of extra security group

thedadams:
If a user did not specify a VPC, then the operator creates one. The
created VPC was also being created with a security group. However, when
the cluster is created, a security group is added automatically.
Therefore, the security group created with the VPC is extraneous.

Issue:
https://github.com/rancher/rancher/issues/29826

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -619,17 +619,15 @@ func (h *Handler) generateAndSetNetworking(svc *cloudformation.CloudFormation, c
 		}
 
 		virtualNetworkString := getParameterValueFromOutput("VpcId", stack.Stacks[0].Outputs)
-		securityGroupsString := getParameterValueFromOutput("SecurityGroups", stack.Stacks[0].Outputs)
 		subnetIdsString := getParameterValueFromOutput("SubnetIds", stack.Stacks[0].Outputs)
 
-		if securityGroupsString == "" || subnetIdsString == "" {
-			return config, fmt.Errorf("no security groups or subnet ids were returned")
+		if subnetIdsString == "" {
+			return config, fmt.Errorf("no subnet ids were returned")
 		}
 
 		config = config.DeepCopy()
 		// copy generated field to status
 		config.Status.VirtualNetwork = virtualNetworkString
-		config.Status.SecurityGroups = strings.Split(securityGroupsString, ",")
 		config.Status.Subnets = strings.Split(subnetIdsString, ",")
 		config.Status.NetworkFieldsSource = "generated"
 	}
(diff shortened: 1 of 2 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -619,17 +619,15 @@ func (h *Handler) generateAndSetNetworking(svc *cloudformation.CloudFormation, c
...
-		if securityGroupsString == "" || subnetIdsString == "" {
-			return config, fmt.Errorf("no security groups or subnet ids were returned")
+		if subnetIdsString == "" {
+			return config, fmt.Errorf("no subnet ids were returned")
 		}
 
 		config = config.DeepCopy()
 		// copy generated field to status
 		config.Status.VirtualNetwork = virtualNetworkString
-		config.Status.SecurityGroups = strings.Split(securityGroupsString, ",")
```
rmweir, line 632: We want to decouple this but not remove it, so this can be removed from this block and the one above it and then set on the status if it is not empty.

rmweir (commented):
We also want to remove the logic in the validator that requires security groups when other networking fields are provided.

controller/eks-cluster-config-handler.go
thedadams: Security groups need to be created with a VPC ID. Therefore, if the user does not provide a customer VPC and Rancher creates one for them, then they cannot provide any security groups. So the security groups will always be empty if Rancher creates the VPC for them.

controller/eks-cluster-config-handler.go
```diff
@@ -878,7 +876,7 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
 			&eks.UpdateClusterConfigInput{
 				Name: aws.String(config.Spec.DisplayName),
 				ResourcesVpcConfig: &eks.VpcConfigRequest{
-					EndpointPublicAccess: config.Spec.PublicAccess,
+					EndpointPublicAccess:  config.Spec.PublicAccess,
```
rmweir, line 879: nit: extra whitespace?

rmweir (approved):
LGTM with 1 nit

controller/eks-cluster-config-handler.go
cbron: Then this means users can't add additional security groups if they don't bring a custom VPC. I see that requirement was crossed out in original issue description now which is good. We may need to document this somewhere.

controller/eks-cluster-config-handler.go
cbron: bump

thedadams: This was a `go fmt` thing so that value of the parameters line up visually. Is it important that I revert this?

cbron: na

cbron (approved):

PR #22, opened 2020-11-05
Use GPU instance type when selected

thedadams:
There was an omission that didn't include GPU instances if a user selected it in the UI.

Issue:
https://github.com/rancher/rancher/issues/29949

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -1326,6 +1326,12 @@ func createNodeGroup(eksConfig *eksv1.EKSClusterConfig, group eksv1.NodeGroup, e
 		},
 	}
 
+	if gpu := group.Gpu; aws.BoolValue(gpu) {
+		nodeGroupCreateInput.AmiType = aws.String(eks.AMITypesAl2X8664Gpu)
+	} else {
+		nodeGroupCreateInput.AmiType = aws.String(eks.AMITypesAl2X8664)
+	}
+
 	if sshKey := group.Ec2SshKey; aws.StringValue(sshKey) != "" {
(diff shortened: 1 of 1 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -1326,6 +1326,12 @@ func createNodeGroup(eksConfig *eksv1.EKSClusterConfig, group eksv1.NodeGroup, e
 		},
 	}
 
+	if gpu := group.Gpu; aws.BoolValue(gpu) {
+		nodeGroupCreateInput.AmiType = aws.String(eks.AMITypesAl2X8664Gpu)
+	} else {
+		nodeGroupCreateInput.AmiType = aws.String(eks.AMITypesAl2X8664)
+	}
```
nickgerace, line 1333: opinionated nit: do not know how much we care about DRY, but you might be able to reduce this to...

```suggestion
      amiType := eks.AMITypesAl2X8664
	if aws.BoolValue(group.Gpu) {
		amiType = eks.AMITypesAl2X8664Gpu
	}
	nodeGroupCreateInput.AmiType  = aws.String(amiType)
```

Granted, this approach can have a scenario where there's two assignment calls, so it's a tradeoff.

nickgerace (approved):

cbron (approved):
lgtm but wait for Rickys review

controller/eks-cluster-config-handler.go
rmweir, line 1333: nit: this is probably actually the default since we never set it before but I think it doesn't hurt to do this

rmweir (approved):
LGTM

PR #25, opened 2020-11-06
Fix gpu field going to `nil`

thedadams:
If the user did not select GPU images, then, during updating the upstream spec, the gpu field would be set to `nil`. This is incorrect because `nil` fields mean that field is not managed by Rancher, which is wrong in this case.

Issue:
https://github.com/rancher/rancher/issues/29949

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -784,6 +784,8 @@ func BuildUpstreamClusterState(name string, clusterState *eks.DescribeClusterOut
 		}
 		if aws.StringValue(ng.Nodegroup.AmiType) == eks.AMITypesAl2X8664Gpu {
 			ngToAdd.Gpu = aws.Bool(true)
+		} else if aws.StringValue(ng.Nodegroup.AmiType) == eks.AMITypesAl2X8664 {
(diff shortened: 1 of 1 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -784,6 +784,8 @@ func BuildUpstreamClusterState(name string, clusterState *eks.DescribeClusterOut
 		}
 		if aws.StringValue(ng.Nodegroup.AmiType) == eks.AMITypesAl2X8664Gpu {
 			ngToAdd.Gpu = aws.Bool(true)
+		} else {
+			ngToAdd.Gpu = aws.Bool(false)
```
cbron, line 788: I take it this isn’t the default ?

cbron (approved):
Seems fine, would like Ricky to approve

controller/eks-cluster-config-handler.go
rmweir: @cbron It's a pointer so default is nil

rmweir, line 788: Nit: this could all be condensed to one line but this is good too.

rmweir (approved):
LGTM

PR #26, opened 2020-11-17
Stop checking name collisions on imported clusters

thedadams:
We don't want to check for name collisions for imported clusters because the imported cluster obviously already exists in EKS.

Issue:
https://github.com/rancher/rancher/issues/29103


aiyengar2 (approved):

rmweir (approved):
LGTM

dramich (approved):

PR #27, opened 2020-12-08
Add support for launch templates

thedadams:
Every time a new cluster is created, the operator creates a new launch
template that will be managed behind the scenes. The launch template is
tagged as managed by Rancher and it should not be modified or deleted.
This launch template is deleted when the cluster is deleted.

If a cluster is imported, then the operator looks for a launch template
with the predefined name. It is used as a managed launch template. When
an imported cluster is deleted, this template is not deleted because the
cluster is not removed from EKS in this case.

For each new node group, the user can use their own launch template. If
they do not provide a launch template, then the operator creates a new
version of the managed template specifically for this new node group. If
a node group is updated, then a new version of the managed launch
template is created and the old version is deleted.

Issues:
https://github.com/rancher/rancher/issues/28803
https://github.com/rancher/rancher/issues/30019
https://github.com/rancher/rancher/issues/30020


controller/eks-cluster-config-handler.go
```diff
@@ -1020,21 +1062,47 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
+			if lt == nil && config.Status.LaunchTemplateID == aws.StringValue(upstreamNg.LaunchTemplate.ID) {
+				// In this case, Rancher is managing the launch template, so we check to see if we need a new version.
+				lt, err = newLaunchTemplateVersionIfNeeded(config, upstreamNg, ng, ec2Service)
 				if err != nil {
 					return config, err
 				}
-				continue
 			}
+
+			if lt != nil && (aws.StringValue(lt.ID) != aws.StringValue(upstreamNg.LaunchTemplate.ID) || aws.Int64Value(lt.Version) != upstreamTemplateVersion) {
```
rmweir, line 1083: I thought the launch template ID could not be change on a nodegroup, is that incorrect?

controller/eks-cluster-config-handler.go
```diff
@@ -1020,21 +1062,47 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
+				attemptUpdatingNodegroups = true
+				ngVersionInput.LaunchTemplate = &eks.LaunchTemplateSpecification{
+					Id:      lt.ID,
+					Version: aws.String(strconv.FormatInt(*lt.Version, 10)),
+				}
+				if upstreamTemplateVersion > 0 {
+					templateVersionsToDelete = append(templateVersionsToDelete, aws.String(strconv.FormatInt(upstreamTemplateVersion, 10)))
+				}
+			}
+		} else if ng.Version != nil {
```
rmweir, line 1093: Doesn't version here refer to kubernetes version? Don't we want to check this regardless of whether a template is used?

rmweir: Took a look at the sdk code. So is the `UpdateNodegroupVersionInput` struct and `UpdateNodegroupVersion` function are now used to upgrade launch template versions or kubernetes version? Why are these operations mutually exclusive? I thought launch templates were specifically concerned with ec2 intsance configuration?

controller/eks-cluster-config-handler.go
```diff
@@ -1151,6 +1225,13 @@ func (h *Handler) importCluster(config *eksv1.EKSClusterConfig, eksService *eks.
 		}
 	}
 
+	launchTemplatesOutput, err := ec2Service.DescribeLaunchTemplates(&ec2.DescribeLaunchTemplatesInput{
+		LaunchTemplateNames: []*string{aws.String(fmt.Sprintf(launchTemplateNameFormat, config.Spec.DisplayName))},
```
rmweir, line 1229: Is this checking to see if it is using a launch template that is managed by Rancher? Could this situation occur? How would the cluster become detached from rancher without being deleted? Or is this for a different use case?

controller/errors.go
```diff
@@ -0,0 +1,47 @@
+package controller
```
rmweir, line 1: +1 for pulling this out, good idea

pkg/apis/eks.cattle.io/v1/types.go
```diff
@@ -52,26 +52,40 @@ type EKSClusterConfigSpec struct {
...
 type EKSClusterConfigStatus struct {
-	Phase          string   `json:"phase"`
-	VirtualNetwork string   `json:"virtualNetwork"`
-	Subnets        []string `json:"subnets"`
-	SecurityGroups []string `json:"securityGroups"`
+	Phase                    string    `json:"phase"`
+	VirtualNetwork           string    `json:"virtualNetwork"`
+	Subnets                  []string  `json:"subnets"`
+	SecurityGroups           []string  `json:"securityGroups"`
+	LaunchTemplateID         string    `json:"launchTemplateID"`
```
rmweir, line 59: Can you name this `generatedLaunchTemplateID` or `managedLaunchTemplateID` to clarify that is the only time it is set?

utils/map.go
```diff
@@ -43,3 +44,42 @@ func GetKeysToDelete(tags map[string]string, upstreamTags map[string]string) []*
...
+	}
+}
+
+func CompareStringMaps(map1, map2 map[string]string) bool {
+	for key, value1 := range map1 {
+		if value2, exists := map2[key]; !exists || value1 != value2 {
+			return false
+		}
+	}
+	return len(map1) == len(map2)
```
rmweir, line 84: nit: can just return false early if this is not true and skip iterating. Probably doesn't actually matter for resource tags since there can only be 50 I believe.

rmweir (changes requested):

controller/errors.go
rmweir, line 1: +1 for pulling this out, good idea

controller/nodegroup.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		logrus.Warnf("could not delete version [%s] of launch template [%s]: %v, will not retry",
+			*launchTemplateDeleteVersionInput.Versions[0],
+			*launchTemplateDeleteVersionInput.LaunchTemplateId,
+			err,
+		)
+	}
+}
+
+func NodeGroupDataFromLaunchTemplates(upstreamSpec *eksv1.EKSClusterConfigSpec, templateID string, ec2Service *ec2.EC2, includeManagedLaunchTemplate bool) error {
```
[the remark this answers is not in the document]
thedadams: Changed to `SetNodeGroupDataFromLaunchTemplates`.

controller/eks-cluster-config-handler.go
```diff
@@ -973,30 +1009,36 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
+			templateVersionsToDelete = append(templateVersionsToDelete, templateVersionToDelete)
+		}
 	}
 
 	if updatingNodegroups {
+		if len(templateVersionsToDelete) != 0 {
+			config = config.DeepCopy()
+			config.Status.Phase = eksConfigUpdatingPhase
+			config.Status.TemplateVersionsToDelete = templateVersionsToDelete
+			return h.eksCC.UpdateStatus(config)
```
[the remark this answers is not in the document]
thedadams: I updated `h.checkAndUpdate` so that it is not possible for this function (`h.updateUpstreamClusterState`) to be called with `Status.TemplateVersionsToDelete` being non-nil.

controller/eks-cluster-config-handler.go
```diff
@@ -324,12 +304,17 @@ func (h *Handler) checkAndUpdate(config *eksv1.EKSClusterConfig, eksService *eks
...
 	}
 
-	upstreamSpec, clusterARN, err := BuildUpstreamClusterState(config.Spec.DisplayName, clusterState, nodeGroupStates, eksService)
+	upstreamSpec, clusterARN, err := BuildUpstreamClusterState(config.Spec.DisplayName, clusterState, nodeGroupStates)
 	if err != nil {
 		return config, err
 	}
 
-	return h.updateUpstreamClusterState(upstreamSpec, config, clusterARN, nodegroupARNs, eksService, svc)
+	err = NodeGroupDataFromLaunchTemplates(upstreamSpec, config.Status.LaunchTemplateID, ec2Service, true)
```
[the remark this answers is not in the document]
thedadams: This is a result of our previous conversation: putting this in `BuildUpstreamClusterState`, then the signature of `BuildUpstreamClusterState` would be much more specific to EKS. You were interested in keeping that function's signature more generic so I pulled out the parts that made the signature more specific to EKS and the launch template implementation.

I am happy to put it back the way it was if you think that was better.

controller/eks-cluster-config-handler.go
thedadams: You can change the launch template ID in EKS.

controller/eks-cluster-config-handler.go
thedadams: I took away the else statement, so `LaunchTemplate` and `Version` can be updated in a single call.

controller/eks-cluster-config-handler.go
thedadams: If the user imports a cluster, creates a new node group (which would create the launch template), removes the cluster from Rancher, and then imports it again: then the launch template would still be there and Rancher wouldn't be able to create a new launch template with the prescribed name.

Therefore, if a launch template with the prescribed name exist on import, the operator will get its ID so it can be managed by Rancher.

pkg/apis/eks.cattle.io/v1/types.go
thedadams: Done.

utils/map.go
thedadams: Done.

controller/eks-cluster-config-handler.go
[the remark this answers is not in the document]
rmweir: I do believe that we may need to fit the function in an interface in the future. Maybe we can add an argument that is more generic such as `params map[string]string` or you can go ahead and pass anything that is simple, ideally a string, and whoever ends up generalizing cloud operators can make this change in EKS at that time. We do want this for syncing right? That is my main concern.

controller/eks-cluster-config-handler.go
rmweir: The problem wasn't whether it did it in a single call, it just looked like it would not upgrade thee kubernetes version if there was a launch template.

controller/eks-cluster-config-handler.go
rmweir: The problem wasn't whether it did it in a single call, it just looked like it would not upgrade thee kubernetes version if there was a launch template.

controller/eks-cluster-config-handler.go
```diff
@@ -1020,23 +1064,51 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
+		}
+
+		if ng.Version != nil {
+			if aws.StringValue(upstreamNg.Version) != desiredNgVersions[aws.StringValue(ng.NodegroupName)] {
+				attemptUpdatingNodegroups = true
+				ngVersionInput.Version = aws.String(desiredNgVersions[aws.StringValue(ng.NodegroupName)])
 			}
 		}
 
+		if ngVersionInput.Version != nil || ngVersionInput.LaunchTemplate != nil {
```
rmweir, line 1104: Have you tested if upgrading both at once works?

rmweir: nit: could move `attemptUpdatingNodegroup = true` here and remove from both blocks

controller/eks-cluster-config-handler.go
rmweir: Okay that makes sense. I suppose we can't clean it up and we still don't want them managing it as their own since we need it for the other nodegroups. This does remind me, is there a reason why we don't make a template per nodegroup?

controller/eks-cluster-config-handler.go
rmweir: Ah, so syncing could cause it. Do we allow users to update the template ID in Rancher?

rmweir (changes requested):

controller/eks-cluster-config-handler.go
rmweir: This looks good though, will resolve pending other question about this logic.

controller/eks-cluster-config-handler.go
thedadams: I believe the main reason was limits: 5,000 launch templates, but 10,000 versions per template. Another reason: it is more straightforward to tell if the launch template is rancher-managed because it will have the correct template id.

controller/eks-cluster-config-handler.go
thedadams: Yes, updating the launch template and version simultaneously works. Also move the `attemptUpdatingNodegroup = true` as you suggested.

controller/eks-cluster-config-handler.go
```diff
@@ -166,18 +162,29 @@ func (h *Handler) OnEksConfigRemoved(key string, config *eksv1.EKSClusterConfig)
...
 		time.Sleep(10 * time.Second)
 		logrus.Infof("waiting for config [%s] node groups to delete", config.Name)
 	}
 
+	if config.Status.ManagedLaunchTemplateID != "" {
+		logrus.Infof("deleting common launch template for config [%s]", config.Name)
+		_, err = ec2Service.DeleteLaunchTemplate(&ec2.DeleteLaunchTemplateInput{
+			LaunchTemplateId: aws.String(config.Status.ManagedLaunchTemplateID),
+		})
+		if err != nil {
```
rmweir, line 183: Is there a particular error that tends to happen often? If this is just for not found error then we can check for that specifically. We seem to retry on most other resources.

controller/nodegroup.go
```diff
@@ -0,0 +1,234 @@
...
+
+func deleteLaunchTemplateVersions(templateID string, templateVersions []*string, ec2Service *ec2.EC2) {
+	launchTemplateDeleteVersionInput := &ec2.DeleteLaunchTemplateVersionsInput{
+		LaunchTemplateId: aws.String(templateID),
+		Versions:         templateVersions,
+	}
+
+	_, err := ec2Service.DeleteLaunchTemplateVersions(launchTemplateDeleteVersionInput)
+	if err != nil {
+		logrus.Warnf("could not delete version [%s] of launch template [%s]: %v, will not retry",
```
rmweir, line 119: I understand there are some issues with us returning errors here. Even if we only show the errors not related to a race, they would confusee the user and incorrectly imply that an update was not successful. Maybe we can at least retry here 3-5 times. I feel it is important we do our best to clean these up. I do worry that it will not be apparent if something is going on that prevents versions from being deleted so AWS ends up with a ton of them.

rmweir (changes requested):

pkg/apis/eks.cattle.io/v1/types.go
```diff
@@ -52,26 +52,40 @@ type EKSClusterConfigSpec struct {
...
+	MaxSize              *int64             `json:"maxSize"`
+	MinSize              *int64             `json:"minSize"`
+	Subnets              []string           `json:"subnets"`
+	Tags                 map[string]*string `json:"tags"`
+	ResourceTags         map[string]*string `json:"resourceTags"`
+	UserData             *string            `json:"userData"`
+	Version              *string            `json:"version"`
+	LaunchTemplate       *LaunchTemplate    `json:"launchTemplate"`
+	RequestSpotInstances *bool              `json:"requestSpotInstances"`
+	SpotInstanceTypes    []*string          `json:"spotInstanceTypes"`
```
rmweir, line 84: I think for subnets I did []string because slices I wouldn't have to consider nil elements, do you have thoughts on one over the other?

rmweir: My guess is that these probably wouldn't be nil from syncing, just someone sending it that way. This is good because it probably matches AWS API. As long as someone sending nil elements doesn't cause a panic or anything this should be good.

thedadams: Yes, the reason I made `SpotInstanceTypes` a `[]*string` was to match the AWS API. I verified that passing a `nil` value will not panic. However, passing a value of `[]*string{nil}` will fail to create the node group, but I would classify that as a user error.

controller/nodegroup.go
thedadams: Implemented a retry that will check for "no found" errors and retry deleting of template versions that don't have "not found" errors.

controller/eks-cluster-config-handler.go
thedadams: Implemented a retry that will check for "no found" errors and retry deleting of templates that don't have "not found" errors.

controller/eks-cluster-config-handler.go
```diff
@@ -1151,6 +1238,13 @@ func (h *Handler) importCluster(config *eksv1.EKSClusterConfig, eksService *eks.
 		}
 	}
 
+	launchTemplatesOutput, err := ec2Service.DescribeLaunchTemplates(&ec2.DescribeLaunchTemplatesInput{
+		LaunchTemplateNames: []*string{aws.String(fmt.Sprintf(launchTemplateNameFormat, config.Spec.DisplayName))},
```
rmweir, line 1242: could we actually somehow tag/label the nodegroups so we know Rancher owns them instead of checking the name? or do both? tagging the resource could also be used for the user or aws admin.

rmweir (changes requested):
LGTM, should include any reserved launch template names and how eks-operator assumes their ownership of them.

controller/eks-cluster-config-handler.go
thedadams: This is looking for the launch template with the prescribed name, not the nodegroup.

We do also tag the launch template when it is created. The problem in this case is: if a launch template with this name already exists, then Rancher won't be able to create one. Therefore, Rancher takes over ownership of the launch template so it can create new versions of it.

rmweir: Sounds good.

rmweir (approved):
LGTM, should include any reserved launch template names and how eks-operator assumes their ownership of them in docs PR.

controller/eks-cluster-config-handler.go
```diff
@@ -578,6 +555,27 @@ func (h *Handler) validateCreate(config *eksv1.EKSClusterConfig, eksService *eks
...
 		cannotBeNilError := "field [%s] cannot be nil for nodegroup [%s] in non-nil cluster [%s]"
 		if !config.Spec.Imported {
+			if ng.LaunchTemplate != nil {
+				if ng.LaunchTemplate.ID == nil {
+					return fmt.Errorf(cannotBeNilError, "launchTemplate.ID", *ng.NodegroupName, config.Name)
+				}
+				if ng.LaunchTemplate.Version == nil {
+					return fmt.Errorf(cannotBeNilError, "launchTemplate.Version", *ng.NodegroupName, config.Name)
+				}
+			} else {
```
nickgerace, line 565: Is this else block needed? Do you want to check if the fields are not nil regardless of the conditional?

controller/eks-cluster-config-handler.go
```diff
@@ -793,27 +792,69 @@ func BuildUpstreamClusterState(name string, clusterState *eks.DescribeClusterOut
...
+		}
+
+		if ng.Nodegroup.LaunchTemplate != nil {
+			if managedTemplateID == aws.StringValue(ng.Nodegroup.LaunchTemplate.Id) {
+				// If this is a rancher-managed launch template, then we move the data from the launch template to the node group.
+				launchTemplateRequestOutput, err := ec2Service.DescribeLaunchTemplateVersions(&ec2.DescribeLaunchTemplateVersionsInput{
+					LaunchTemplateId: ng.Nodegroup.LaunchTemplate.Id,
+					Versions:         []*string{ng.Nodegroup.LaunchTemplate.Version},
+				})
+				if err != nil || len(launchTemplateRequestOutput.LaunchTemplateVersions) == 0 {
```
nickgerace, line 818: Do we want to return or report the error in some way here? It's ignored currently, which may be the right idea, but wanted to make sure.

controller/eks-cluster-config-handler.go
```diff
@@ -793,27 +792,69 @@ func BuildUpstreamClusterState(name string, clusterState *eks.DescribeClusterOut
...
+					versionNumber, err := strconv.ParseInt(aws.StringValue(ng.Nodegroup.LaunchTemplate.Version), 10, 64)
+					if err == nil {
+						version = aws.Int64(versionNumber)
+					}
+
+					ngToAdd.LaunchTemplate = &eksv1.LaunchTemplate{
+						ID:      ng.Nodegroup.LaunchTemplate.Id,
+						Name:    ng.Nodegroup.LaunchTemplate.Name,
+						Version: version,
+					}
```
nickgerace, line 841: I think if an error occurs we have an uninitialized version.

pkg/apis/eks.cattle.io/v1/types.go
```diff
@@ -52,26 +52,40 @@ type EKSClusterConfigSpec struct {
...
-	Phase          string   `json:"phase"`
-	VirtualNetwork string   `json:"virtualNetwork"`
-	Subnets        []string `json:"subnets"`
-	SecurityGroups []string `json:"securityGroups"`
+	Phase                    string    `json:"phase"`
+	VirtualNetwork           string    `json:"virtualNetwork"`
+	Subnets                  []string  `json:"subnets"`
+	SecurityGroups           []string  `json:"securityGroups"`
+	ManagedLaunchTemplateID  string    `json:"managedLaunchTemplateID"`
+	TemplateVersionsToDelete []*string `json:"templateVersionsToDelete"`
```
cbron, line 60: Can you elaborate on this a bit, why does it happen afterwards. Is there a chance for a race condition between when the new one is added and old one is deleted in checkAndUpdate ?

controller/eks-cluster-config-handler.go
```diff
@@ -1020,21 +1074,48 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
 		// happen together
 
 		ng := ngs[aws.StringValue(upstreamNg.NodegroupName)]
```
cbron, line 1076: Did you test updating the launch template from both sides ? The CR and the AWS console.

controller/eks-cluster-config-handler.go
```diff
@@ -65,7 +65,7 @@ func Register(
 	eks.OnRemove(ctx, controllerRemoveName, controller.OnEksConfigRemoved)
 }
 
-func (h *Handler) OnEksConfigChanged(key string, config *eksv1.EKSClusterConfig) (*eksv1.EKSClusterConfig, error) {
+func (h *Handler) OnEksConfigChanged(_ string, config *eksv1.EKSClusterConfig) (*eksv1.EKSClusterConfig, error) {
```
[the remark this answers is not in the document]
thedadams: The operator implements an interface, so we need the correct signature.

controller/eks-cluster-config-handler.go
thedadams: If the node group has a launch template, then the values in the `else` block are taken care of in the launch template. If the node group doesn't have a launch template, then we need to check these values.

controller/eks-cluster-config-handler.go
thedadams: Yes, this has changed in the next update to this PR.

controller/eks-cluster-config-handler.go
thedadams: Then EKS will just use the default version, which is the desired outcome.

controller/eks-cluster-config-handler.go
```diff
@@ -793,27 +792,69 @@ func BuildUpstreamClusterState(name string, clusterState *eks.DescribeClusterOut
...
+						ID:      ng.Nodegroup.LaunchTemplate.Id,
+						Name:    ng.Nodegroup.LaunchTemplate.Name,
+						Version: version,
+					}
+				}
+			}
+		} else {
+			// If the node group does not have a launch template, then the following must be pulled from the node group config.
+			if !aws.BoolValue(ngToAdd.RequestSpotInstances) && len(ng.Nodegroup.InstanceTypes) > 0 {
+				ngToAdd.InstanceType = ng.Nodegroup.InstanceTypes[0]
```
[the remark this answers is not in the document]
thedadams: This was a choice that Ricky made on first implementation. A nodeGroup can have more than once instance type.

controller/nodegroup.go
```diff
@@ -0,0 +1,274 @@
...
+			return templateVersionToDelete, false, err
+		}
+
+		if ngState.Nodegroup.LaunchTemplate != nil &&
+			aws.StringValue(ngState.Nodegroup.LaunchTemplate.Id) == config.Status.ManagedLaunchTemplateID {
+			templateVersionToDelete = ngState.Nodegroup.LaunchTemplate.Version
+		}
+	}
+
+	return templateVersionToDelete, true, err
```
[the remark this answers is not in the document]
thedadams: No, the versions only get deleted after the node group has been successfully deleted.

controller/nodegroup.go
```diff
@@ -0,0 +1,274 @@
...
+const (
+	launchTemplateNameFormat = "rancher-managed-lt-%s"
+	launchTemplateTagKey     = "rancher-managed-template"
+	launchTemplateTagValue   = "do-not-modify-or-delete"
+	storageDeviceName        = "/dev/sdr"
+)
+
+func createLaunchTemplate(clusterDisplayName string, ec2Service *ec2.EC2) (*eksv1.LaunchTemplate, error) {
+	launchTemplateCreateInput := &ec2.CreateLaunchTemplateInput{
+		LaunchTemplateData: &ec2.RequestLaunchTemplateData{UserData: aws.String("placeholder")},
```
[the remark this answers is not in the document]
thedadams: We don't want to use the default version (1) of the rancher-managed launch template because it can never be deleted (until the launch template is deleted). Therefore, we create the launch template with version 1 having this placeholder userdata because it can't be blank.

controller/nodegroup.go
```diff
@@ -0,0 +1,274 @@
...
+		aws.StringValue(upstreamNg.Ec2SshKey) != aws.StringValue(ng.Ec2SshKey) ||
+		aws.Int64Value(upstreamNg.DiskSize) != aws.Int64Value(ng.DiskSize) ||
+		aws.StringValue(upstreamNg.ImageID) != aws.StringValue(ng.ImageID) ||
+		(!aws.BoolValue(upstreamNg.RequestSpotInstances) && aws.StringValue(upstreamNg.InstanceType) != aws.StringValue(ng.InstanceType)) ||
+		!utils.CompareStringMaps(aws.StringValueMap(upstreamNg.ResourceTags), aws.StringValueMap(ng.ResourceTags)) {
+		lt, err := createNewLaunchTemplateVersion(config.Status.ManagedLaunchTemplateID, ng, ec2Service)
+		if err != nil {
+			return nil, err
+		}
+		config.Status.TemplateVersionsToDelete = append(config.Status.TemplateVersionsToDelete, aws.String(strconv.FormatInt(*upstreamNg.LaunchTemplate.Version, 10)))
```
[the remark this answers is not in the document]
thedadams: Yes, after the node group finishes its update and the cluster goes back to active status.

pkg/apis/eks.cattle.io/v1/types.go
thedadams: If the launch template version is deleted before the node group goes into updating status, then EKS will not allow the node group to be updated. EKS will put the node group in a bad status and the only recovery is to delete the node group and create a new one (or visa versa).

controller/eks-cluster-config-handler.go
thedadams: Yep.

controller/eks-cluster-config-handler.go
nickgerace: What happens if `if err == nil` is false? Isn't `version` uninitialized at that point, and then `Version: version` will be unset?

thedadams: Yes, and EKS will use the default version of the template, which is the desired result.

nickgerace (approved):
LGTM

controller/nodegroup.go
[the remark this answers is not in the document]
cbron: Surprised user data can't be blank. So we add a placeholder string, and that would be on every version 1 of every instance-template ? Or only those where user doesn't specify one.

cbron (approved):
Last thing is the user-data piece, which seems ok, I'd just like to understand what is fully going on and if there are any negative side effects of that approach. Aside from that, I don't fully grep everything here as I am not as familiar with eks-operator, but what I do seems right.

controller/nodegroup.go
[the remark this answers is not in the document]
thedadams: When I said, "it can't be blank" I meant the launch template. We want to create a launch template, but we don't want to use version 1 because it can never be deleted (and we want to be able to clean up versions when node groups are deleted/updated). Therefore this first version will not be used. Since we can't create a blank launch template, we create one with fake userdata.

I added a comment in the code to explain this as well.

controller/errors.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return false
+}
+
+// NodeGroupIssueIsUpdatable checks to see the node group can be updated with the given issue code.
+func NodeGroupIssueIsUpdatable(code string) bool {
+	return code == eks.NodegroupIssueCodeAsgInstanceLaunchFailures ||
+		code == eks.NodegroupIssueCodeInstanceLimitExceeded ||
+		code == eks.NodegroupIssueCodeInsufficientFreeAddresses ||
+		code == eks.NodegroupIssueCodeClusterUnreachable
```
rmweir, line 61: just making sure that these are the errors that can be resolved?

thedadams: Yes, these errors cannot be recovered by updating.

rmweir (approved):
LGTM

controller/eks-cluster-config-handler.go
```diff
@@ -1041,15 +1052,16 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
 			return config, err
 		}
+		templateVersionsToAdd[aws.StringValue(ng.NodegroupName)] = ltVersion
 		updatingNodegroups = true
 	}
 
-	// check if node groups need to be deleted
-	templateVersionsToDelete := make([]*string, 0)
+	// check for node groups need to be deleted
+	templateVersionsToDelete := make(map[string]string)
```
rmweir, line 1064: 👍

controller/eks-cluster-config-handler.go
```diff
@@ -1087,6 +1101,7 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
 	}
 
 	var attemptUpdatingNodegroups bool
```
rmweir, line 1103: I think it may be confusing that there are two separate chunks of updating node groups logic, but I believe one is creating/deleting node groups and the other is updating node group properties. Can we name `attemptUpdatingNodegroups` to something `updateNodegroupProperties` so it's clear what the two are for? If I'm misunderstanding please clarify.

rmweir (changes requested):

controller/eks-cluster-config-handler.go
thedadams: Done

rmweir (approved):
LGTM

PR #29, opened 2021-01-27
Add helm annotation and required nodegroup field

thedadams:
Add a helm annotation to the the generated CRD. Also, add `nodegoupName` as a required field on the `NodeGroup` type.

Issue:
https://github.com/rancher/rancher/issues/30997

### DIFF
--- a/pkg/apis/eks.cattle.io/v1/types.go
+++ b/pkg/apis/eks.cattle.io/v1/types.go
@@ -67,9 +67,9 @@ type EKSClusterConfigStatus struct {
 type NodeGroup struct {
 	Gpu                  *bool              `json:"gpu"`
 	ImageID              *string            `json:"imageId"`
-	NodegroupName        *string            `json:"nodegroupName" norman:"required"`
+	NodegroupName        *string            `json:"nodegroupName" norman:"required" wrangler:"required"`
 	DiskSize             *int64             `json:"diskSize"`
-	InstanceType         *string            `json:"instanceType" norman:"required"`
+	InstanceType         *string            `json:"instanceType"`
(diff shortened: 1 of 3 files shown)


pkg/apis/eks.cattle.io/v1/types.go
```diff
@@ -67,9 +67,9 @@ type EKSClusterConfigStatus struct {
 type NodeGroup struct {
 	Gpu                  *bool              `json:"gpu"`
 	ImageID              *string            `json:"imageId"`
-	NodegroupName        *string            `json:"nodegroupName" norman:"required"`
+	NodegroupName        *string            `json:"nodegroupName" norman:"required" wrangler:"required"`
```
cbron, line 70: I haven't seen wrangler required before. Is that needed for codegen ?

thedadams: Yes. The codegen generates the CRD from the type. It looks for different struct tags and adds information to the CRD based on the tags. In this case, this adds `nodegroupName` to the `required` field for the `NodeGroups` (also in the diff).

cbron (approved):
lgtm, but wait for @rmweir  to approve.

pkg/apis/eks.cattle.io/v1/types.go
rmweir: Really? I thought it recognized the norman tags?

thedadams: Running `go generate` with just the norman tags did not produce the `required` field.

rmweir (approved):
LGTM

PR #30, opened 2021-02-08
Correctly identify when GPU instance should be used

thedadams:
ImageId was (correctly) passed as an empty string instead of null, so
eks-operator was not looking to see if a GPU instance was requested.

Issue:
https://github.com/rancher/rancher/issues/31147

### DIFF
--- a/controller/nodegroup.go
+++ b/controller/nodegroup.go
@@ -219,7 +219,10 @@ func createNodeGroup(config *eksv1.EKSClusterConfig, group eksv1.NodeGroup, eksS
 		}
 	}
 
-	launchTemplateVersion := aws.String(strconv.FormatInt(*lt.Version, 10))
+	var launchTemplateVersion *string
+	if aws.Int64Value(lt.Version) != 0 {
+		launchTemplateVersion = aws.String(strconv.FormatInt(*lt.Version, 10))
+	}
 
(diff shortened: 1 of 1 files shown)


controller/nodegroup.go
```diff
@@ -219,7 +219,10 @@ func createNodeGroup(config *eksv1.EKSClusterConfig, group eksv1.NodeGroup, eksS
 		}
 	}
 
-	launchTemplateVersion := aws.String(strconv.FormatInt(*lt.Version, 10))
+	var launchTemplateVersion *string
+	if lt.Version != nil {
```
rmweir, line 223: When would this be nil?

thedadams: It would only be `nil` if it was passed that way, but a better check has been added.

rmweir (approved):
LGTM

controller/nodegroup.go
```diff
@@ -230,7 +233,7 @@ func createNodeGroup(config *eksv1.EKSClusterConfig, group eksv1.NodeGroup, eksS
 		nodeGroupCreateInput.InstanceTypes = group.SpotInstanceTypes
 	}
 
-	if group.ImageID == nil {
+	if aws.StringValue(group.ImageID) == "" {
```
cbron, line 236: Confirmed this would return "" for nil, so looks good.

cbron (approved):

PR #31, opened 2021-02-11
Delete launch template versions on error

thedadams:
If there is an error creating or updating a node group with a
managed launch template, then there could be many versions of the
launch template created. The operator should clean them up so there are
not a bunch leftover for no reason.

Issue:
https://github.com/rancher/rancher/issues/31263

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -1127,6 +1127,9 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
 				}
 
 				if lt != nil {
+					if upstreamTemplateVersion > 0 {
+						templateVersionsToDelete[aws.StringValue(upstreamNg.NodegroupName)] = strconv.FormatInt(upstreamTemplateVersion, 10)
+					}
 					templateVersionsToAdd[aws.StringValue(ng.NodegroupName)] = strconv.FormatInt(*lt.Version, 10)
 				}
 			}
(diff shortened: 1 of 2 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -1152,6 +1152,11 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
 			updateNodegroupProperties = true
 			_, err := eksService.UpdateNodegroupVersion(ngVersionInput)
 			if err != nil {
+				if version, ok := templateVersionsToAdd[aws.StringValue(ng.NodegroupName)]; ok {
+					// If there was an error updating the node group and a Rancher-managed launch template version was created,
+					// then the version that caused the issue needs to be delete to prevent bad versions from piling up.
+					deleteLaunchTemplateVersions(config.Status.ManagedLaunchTemplateID, []*string{aws.String(version)}, ec2Service)
+				}
 				return config, err
```
rmweir, line 1160: This made me realize something, isn't it possible for config's status to be change in `newLaunchTemplateVersionIfNeeded` but never actually updated? In this case we wouldn't want it updated any way I think, but I'm guessing this messes up the cache. Also, seem like a side effect of the function that should maybe be brought out so it's easier to track.

rmweir (changes requested):

controller/eks-cluster-config-handler.go
thedadams: Yes, you're right. That absolutely needs to be pulled out. Will update.

rmweir (approved):
LGTM

cbron (approved):

PR #32, opened 2021-02-17
Use root device name when setting disk size

thedadams:
The launch template data should use the image's root device name when
setting the disk size. Otherwise, a new device is attached to the image,
which is not the intended outcome. The root device size should be set to
the size the user submits.

Issue:
https://github.com/rancher/rancher/issues/31361

### DIFF
--- a/controller/nodegroup.go
+++ b/controller/nodegroup.go
@@ -21,7 +21,7 @@ const (
 	launchTemplateNameFormat = "rancher-managed-lt-%s"
 	launchTemplateTagKey     = "rancher-managed-template"
 	launchTemplateTagValue   = "do-not-modify-or-delete"
-	storageDeviceName        = "/dev/sdr"
+	defaultStorageDeviceName = "/dev/xvda"
 )
 
 func createLaunchTemplate(clusterDisplayName string, ec2Service *ec2.EC2) (*eksv1.LaunchTemplate, error) {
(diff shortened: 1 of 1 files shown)


rmweir (approved):
LGTM, I would just make sure that there is not any kind of default storage device variable in the sdk before we make our own.

thedadams:
Nothing in the SDK about a default device name. I use it because that is the root device name of the Amazon EKS Linux image.

controller/nodegroup.go
```diff
@@ -306,3 +315,15 @@ func deleteNodeGroup(config *eksv1.EKSClusterConfig, ng eksv1.NodeGroup, eksServ
...
 	return templateVersionToDelete, true, err
 }
+
+func getImageRootDeviceName(imageID *string, ec2Service *ec2.EC2) (*string, error) {
+	describeOutput, err := ec2Service.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{imageID}})
+	if err != nil {
+		return nil, err
+	}
+	if len(describeOutput.Images) == 0 {
+		return nil, fmt.Errorf("no images returned for id %v", aws.StringValue(imageID))
```
cbron, line 325: nit: could say "Cannot describe images for id"....

cbron (approved):

PR #34, opened 2021-03-11
Add build for arm64

thedadams:
It is important for eks-operator to support arm64. Here, the support is
added through introducing dapper support.

Issue:
https://github.com/rancher/rancher/issues/31517


.drone.yml
```diff
@@ -1,80 +1,132 @@
...
+    path: /var/run/docker.sock
   commands:
-  - docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
-  - TAG=${DRONE_TAG} make image-push
-  environment:
-    DOCKER_USERNAME:
-      from_secret: docker_username
-    DOCKER_PASSWORD:
-      from_secret: docker_password
+  - TAG=${DRONE_TAG} dapper image-scan
```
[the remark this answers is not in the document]
rmweir: Right, pretty sure Trivy doesn't throw a non-zero exit code if vulnerabilities are found. Although with a little more logic we could make it functional though. I guess if Trivy couldn't scan it for some reason it'd fail the build but idk how valuable that is.

package/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM alpine
```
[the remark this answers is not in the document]
cbron: Lets stick with UBI

.drone.yml
```diff
@@ -1,80 +1,132 @@
...
+- name: docker-publish
+  image: plugins/docker
+  settings:
+    dockerfile: package/Dockerfile
+    username:
+      from_secret: docker_username
+    password:
+      from_secret: docker_password
+    repo: "rancher/eks-operator"
+    tag: "${DRONE_TAG}-amd64"
```
rmweir, line 32: @cbron with these tags, is the expectation that the user will retag it without the arch so it's compatible with the charts or do we need to add some logic to the charts?

.drone.yml
```diff
@@ -1,80 +1,132 @@
...
 name: linux-amd64
+
 platform:
   os: linux
   arch: amd64
 
 steps:
-- name: ci
-  image: briandowns/rancher-build-base:v0.1.1
+- name: build
```
rmweir, line 10: Looks like this and the arm `ci` pipeline are doing the same thing for diff arch's so I'm confused by the naming. This is doing more than just build right?

Dockerfile.dapper
```diff
@@ -0,0 +1,20 @@
+FROM golang:1.16
+
+ARG DAPPER_HOST_ARCH
+ENV ARCH $DAPPER_HOST_ARCH
+
+RUN apt update && \
+    apt install -y bash git gcc docker.io vim less file curl wget ca-certificates
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.30.0
```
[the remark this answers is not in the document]
rmweir: Why would we need golangci-lint on both arch's anyway? Would it ever give diff results?

scripts/ci
```diff
@@ -1,24 +1,9 @@
...
-fi
-
-echo 'Running: go mod tidy'
-go mod tidy
-if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
-  echo 'go mod tidy produced differences'
-  exit 1
-fi
+./build
+./validate
```
rmweir, line 8: Should validate come before build? Why build if validation fails?

rmweir (changes requested):

.drone.yml
cbron: If you tag v1.0.0 this would create v1.0.0-amd64 and v1.0.0-arm64, but manifest will unify them into just 1.0.0. User doesn't tag anything around arch's.

Dockerfile.dapper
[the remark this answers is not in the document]
cbron: In rancher we only run it on amd64 to save time, should give same results. There is actually an arm build so if you set both up it wouldn't hurt other than wasting resources.

.drone.yml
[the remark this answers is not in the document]
thedadams: Removed.

Dockerfile.dapper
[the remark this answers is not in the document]
thedadams: Made this consistent with rancher: validate only runs on amd64.

package/Dockerfile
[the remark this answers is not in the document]
thedadams: Using `ui8/ubi-minimal`.

package/Dockerfile
```diff
@@ -0,0 +1,4 @@
...
+COPY bin/eks-operator /usr/bin/
+USER 1000
```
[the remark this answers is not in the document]
thedadams: Removed `USER` and added `ENV KUBECONFIG /root/.kube/config`.

.drone.yml
thedadams: Changed back to `ci`.

scripts/ci
thedadams: Switched.

thedadams:
Some changes to codegen were needed to support dapper. These changes were tested to ensure they still generate the same files.

Also, I tested these changes in a dev environment to ensure the container still runs as expected.

package/Dockerfile
```diff
@@ -0,0 +1,5 @@
+ARG UBI_IMAGE=registry.access.redhat.com/ubi8/ubi-minimal:latest
+FROM ${UBI_IMAGE}
```
rmweir, line 2: need to run the following to ensure packages are up to date:
```
RUN microdnf update -y && \ 
    rm -rf /var/cache/yum
```

rmweir (changes requested):

package/Dockerfile
thedadams: Done.

package/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+ENV KUBECONFIG /root/.kube/config
```
cbron, line 3: What is the point of this kubeconfig, does that get dynamically set later on ?

thedadams: There is a command line flag that uses the environment variable for connecting to a different Kubernetes cluster.

rmweir (approved):
LGTM

pkg/codegen/main.go
```diff
@@ -15,13 +15,15 @@ import (
...
+	os.Unsetenv("GOPATH")
+
 	controllergen.Run(args.Options{
 		OutputPackage: "github.com/rancher/eks-operator/pkg/generated",
 		Boilerplate:   "pkg/codegen/boilerplate.go.txt",
 		Groups: map[string]args.Group{
 			"eks.cattle.io": {
 				Types: []interface{}{
-					v12.EKSClusterConfig{},
+					"./pkg/apis/eks.cattle.io/v1",
```
cbron, line 26: Are changes to this file right ? Or is this debug ?

cbron (commented):
Just the one comment then we can merge

pkg/codegen/main.go
thedadams: This is the "some changes to support dapper." For some reason, codegen couldn't find the `v12.EKSClusterConfig` type when running codegen in dapper. Changing the type to the directory fixed this issue. Codegen will scan the directory.

I verified that these changes didn't change what was generated.

cbron: Sounds good. Just keep in mind for GKE/AKS.

cbron (approved):

PR #35, opened 2021-03-11
Fix possible conflict when editing during provision

thedadams:
If a user edits a cluster at the exact right (or wrong) time, then the
UpdateStatus call may produce a conflict and will error. When the
controller re-enters the create function, it will try to verify that a
cluster with the same name in EKS does not exist. It will find the one
that was created and error.

Instead, we retry the UpdateStatus call on conflict to protect against
this happening.

Issue:
https://github.com/rancher/rancher/issues/31608

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -25,7 +25,9 @@ import (
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v15 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -495,9 +497,17 @@ func (h *Handler) create(config *eksv1.EKSClusterConfig, sess *session.Session,
...
+		config, err = h.eksCC.Get(config.Namespace, config.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		config.Status.Phase = eksConfigCreatingPhase
+		config.Status.FailureMessage = ""
+		config, err = h.eksCC.UpdateStatus(config)
+		return err
+	})
+	return config, err
```
rmweir, line 510: After reading the description the use of `RetryOnConflict` makes sense. I think tagging the cluster so we know it belongs to rancher may be an alternative but that complicates things since the user can change tags, we would have to make sure that one is preserved and probably just not display it. If we stick with this can you add a comment on top explaining why it is used here? I think this function should be used sparingly for exceptions like this one and the justification should be clear.

rmweir, line 510: After reading the description the use of RetryOnConflict makes sense. I think tagging the cluster so we know it belongs to rancher may be an alternative but that complicates things since the user can change tags, we would have to make sure that one is preserved and probably just not display it. If we stick with this can you add a comment on top explaining why it is used here? I think this function should be used sparingly for exceptions like this one and the justification should be clear.

rmweir (changes requested):

controller/eks-cluster-config-handler.go
thedadams: Added a comment.

rmweir (approved):
LGTM

cbron (approved):
lgtm, need to make kev2 team aware

PR #38, opened 2021-05-14
Use norman:"pointer" for nullable fields

cmurphy:
Opt into new "pointer" field attribute from norman that will convert any
type to a pointer in the generated client code. The generated client code
includes "omitempty" JSON tags, which means that for maps and slices,
unset and empty values are both sent to the server as unset, which
breaks the cluster create validation. With this change, there's now a
difference between a nil value and an empty allocated map or slice in the
client code. Strings can now also use this attribute.

https://github.com/rancher/rancher/issues/32440
https://github.com/rancher/rancher/issues/32553


thedadams (approved):

rmweir (approved):
LGTM

thedadams (approved):

mjura (approved):
LGTM

cbron (approved):

rmweir (approved):
LGTM

PR #41, opened 2021-05-26
Add additional trusted CAs

thedadams:
When Rancher is installed on a cluster behind a proxy, the operator
should trust the proxy's certificate. It is passed to the deployment via
 a secret the user should setup per the Rancher documentation.

Issue:
https://github.com/rancher/rancher/issues/31846


cmurphy (approved):

charts/eks-operator/templates/deployment.yaml
```diff
@@ -25,3 +25,15 @@ spec:
           value: {{ .Values.httpsProxy }}
         - name: NO_PROXY
           value: {{ .Values.noProxy }}
+{{- if .Values.additionalTrustedCAs }}
+        volumeMounts:
+          - mountPath: /etc/ssl/certs/ca-additional.pem
+            name: tls-ca-additional-volume
+            subPath: ca-additional.pem
```
jiaqiluo, line 32: The current code will result in creating a file at this full path: `/etc/ssl/certs/ca-additional.pem/ca-additional.pem`

```suggestion
          - mountPath: /etc/ssl/certs
            name: tls-ca-additional-volume
```

.gitignore
```diff
@@ -1,4 +1,5 @@
 .idea/
 .vscode/
 bin/
-dist/
\ No newline at end of file
+dist/
+.dapper
```
jiaqiluo, line 5: nit. 
```suggestion
.dapper

```

jiaqiluo (changes requested):

charts/eks-operator/templates/deployment.yaml
thedadams: I think the desired result is to create the file.

jiaqiluo (approved):
LGTM

charts/eks-operator/templates/deployment.yaml
jiaqiluo: this is not an issue, see
https://github.com/rancher/aks-operator/pull/18/files#r641063779

charts/eks-operator/values.yaml
```diff
@@ -9,3 +9,4 @@ eksOperator:
 httpProxy: ""
 httpsProxy: ""
 noProxy: ""
+additionalTrustedCAs: false
```
cbron, line 12: Should we document this in app-readme or readme of chart ? cc @rmweir

cbron (approved):

charts/eks-operator/values.yaml
cbron: Oh nvm, forgot how this worked.

rmweir (approved):
LGTM

charts/eks-operator/values.yaml
rmweir: @cbron what do you mean? I think putting it in a readme is a good idea.

PR #42, opened 2021-06-15
Add USER directive to Docker image

thedadams:
In a hardened cluster, containers must not run as root user.
 Adding the USER directive changes the user to a non-root user, which
 allows the container to run in such environments.

Issues:
https://github.com/rancher/rancher/issues/29066
https://github.com/rancher/rancher/issues/33172

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -4,4 +4,5 @@ ENV KUBECONFIG /root/.kube/config
 RUN microdnf update -y && \
     rm -rf /var/cache/yum
 COPY bin/eks-operator /usr/bin/
+USER 1001
 ENTRYPOINT ["eks-operator"]


jiaqiluo (approved):
LGTM

cmurphy (commented):
Is there a reason to use a hardcoded uid vs running `adduser` and using a named user? I found https://sysdig.com/blog/dockerfile-best-practices/#1-2 recommends against hardcoding the uid, and this https://snyk.io/blog/10-docker-image-security-best-practices/ also recommends using `adduser`.

cmurphy (approved):
Discussed offline, the UBI image doesn't have `adduser`

package/Dockerfile
```diff
@@ -4,4 +4,5 @@ ENV KUBECONFIG /root/.kube/config
 RUN microdnf update -y && \
     rm -rf /var/cache/yum
 COPY bin/eks-operator /usr/bin/
+USER 1001
 ENTRYPOINT ["eks-operator"]
```
cbron, line 8: Don't you also need a PSP that specifies user isn't root ? 
https://github.com/rancher/backup-restore-operator/pull/99/files#diff-5218b12eb3a3217f30883761568497dbdeba7df6fc347fb3869c00cc04895af6R13

And speaking of that you would also need a networkPolicy, and potentially turning off default SA. Is that all coming in follow up PR ?

thedadams: We don't need a PSP for this. Kubernetes knows that the image will not run with root user after making this change.

If I understand correctly, default service accounts and network policies are applied to namespaces. None of the KEv2 operators use their own namespaces. The KEv2 operators do create their own service accounts and don't use the default in the cattle-system namespace.

cbron: Oh, so this is taken care of for us because its already setup that way when we setup rancher. Same for PSP's, see `default-psp-rolebinding`.

cbron (approved):

PR #44, opened 2021-06-22
Use same SLES image as Rancher

thedadams:


cmurphy (approved):

rmweir (approved):

cbron (approved):

cbron:
@thedadams make sure these are linked to an issue where QA would run a test on them, to ensure they still work. You may need to create one for gke/eks.

thedadams:
https://github.com/rancher/rancher/issues/33371

cmurphy (approved):

rmweir (approved):
LGTM

.drone.yml
```diff
@@ -114,6 +125,17 @@ steps:
...
+  image: plugins/docker
+  settings:
+    dockerfile: package/Dockerfile
+    repo: "rancher/eks-operator"
+    tag: "${DRONE_COMMIT}-arm64"
+    dry_run: true
+  when:
+    event:
+    - pull_request
+
```
cbron, line 138: whats going on in these two ? We removed image build from CI so it gets run on PR now ? Whats reasoning for that. And whats difference between this block and block above.

thedadams: There was a lot of discussion on the installation of Docker in the Dapper image because we were using an old version Docker. See https://github.com/rancher/rancher/blob/master/Dockerfile.dapper#L35-L38 for what it looked like.

The outcome of that was instead of building the image in `scripts/package`, let the plugin handle it. So I added a Docker plugin step that builds the image on pull requests, but doesn't push it (no username nor password is added and `push` is set to `false`).

cbron (approved):

PR #45, opened 2021-06-23
Fix label updates

rmweir:
**Problem:**
Prior, if only a label update was necessary, the scaling config
fields were not included in the nodegroup config update. This
is considered invalid and will cause an error to be returned.

**Solution:**
Now, scaling config fields are included whether they need to be
updated or not. An additional bug has been fixed where if some
labels needed to be deleted and other updated, only the updated
labels would be added to the config update.

**Issue:**
https://github.com/rancher/rancher/issues/33302

### DIFF
--- a/controller/nodegroup.go
+++ b/controller/nodegroup.go
@@ -327,3 +327,50 @@ func getImageRootDeviceName(imageID *string, ec2Service *ec2.EC2) (*string, erro
 
 	return describeOutput.Images[0].RootDeviceName, nil
 }
+
+// getNodegroupConfigUpdate returns an UpdateNodegroupConfigInput that represents desired state and a bool
+// indicating whether an update needs to take place to achieve the desired state.
+func getNodegroupConfigUpdate(clusterName string, ng eksv1.NodeGroup, upstreamNg eksv1.NodeGroup) (eks.UpdateNodegroupConfigInput, bool) {
+	nodegroupConfig := eks.UpdateNodegroupConfigInput{
+		ClusterName:   aws.String(clusterName),
+		NodegroupName: ng.NodegroupName,
+		ScalingConfig: &eks.NodegroupScalingConfig{},
+	}
+	var sendUpdateNodegroupConfig bool
+
+	if ng.Labels != nil {
+		unlabels := utils.GetKeysToDelete(aws.StringValueMap(ng.Labels), aws.StringValueMap(upstreamNg.Labels))
+		labels := utils.GetKeyValuesToUpdate(aws.StringValueMap(ng.Labels), aws.StringValueMap(upstreamNg.Labels))
+
+		if unlabels != nil || labels != nil {
+			sendUpdateNodegroupConfig = true
+			nodegroupConfig.Labels = &eks.UpdateLabelsPayload{
+				RemoveLabels:      unlabels,
+				AddOrUpdateLabels: labels,
+			}
+		}
+	}
+
+	if ng.DesiredSize != nil {
+		nodegroupConfig.ScalingConfig.DesiredSize = ng.DesiredSize
+		if aws.Int64Value(upstreamNg.DesiredSize) != aws.Int64Value(ng.DesiredSize) {
+			sendUpdateNodegroupConfig = true
+		}
+	}
+
+	if ng.MinSize != nil {
+		nodegroupConfig.ScalingConfig.MinSize = ng.MinSize
+		if aws.Int64Value(upstreamNg.MinSize) != aws.Int64Value(ng.MinSize) {
+			sendUpdateNodegroupConfig = true
+		}
+	}
+
+	if ng.MaxSize != nil {
+		nodegroupConfig.ScalingConfig.MaxSize = ng.MaxSize
+		if aws.Int64Value(upstreamNg.MaxSize) != aws.Int64Value(ng.MaxSize) {
+			sendUpdateNodegroupConfig = true
+		}
+	}
+
+	return nodegroupConfig, sendUpdateNodegroupConfig
+}
--- a/controller/nodegroup_test.go
+++ b/controller/nodegroup_test.go
@@ -1,0 +1,1 @@
+package controller
--- a/controller/nodegroup_test.go
+++ b/controller/nodegroup_test.go
@@ -1,1 +1,138 @@
 package controller
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/eks"
+	v1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetNodegroupConfigUpdate(t *testing.T) {
+	type nodegroupUpdateTestCase struct {
+		clusterName           string
+		ng1                   v1.NodeGroup
+		ng2                   v1.NodeGroup
+		expectedNgUpdateInput eks.UpdateNodegroupConfigInput
+		expectedNgNeedsUpdate bool
+	}
+	asserts := assert.New(t)
+	testCases := []nodegroupUpdateTestCase{
+		{
+			// test case where there should be no update
+			clusterName: "testcluster1",
+			ng1:         v1.NodeGroup{Labels: aws.StringMap(map[string]string{"a": "b"}), MinSize: aws.Int64(1), MaxSize: aws.Int64(1)},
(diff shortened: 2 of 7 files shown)


controller/nodegroup.go
```diff
@@ -327,3 +327,50 @@ func getImageRootDeviceName(imageID *string, ec2Service *ec2.EC2) (*string, erro
 
 	return describeOutput.Images[0].RootDeviceName, nil
 }
+
+// getNodegroupConfigUpdate returns an UpdateNodegroupConfigInput that represents desired state and a bool
+// indicating whether an update needs to take place to achieve the desired state.
+func getNodegroupConfigUpdate(clusterName string, ng eksv1.NodeGroup, upstreamNg eksv1.NodeGroup) (eks.UpdateNodegroupConfigInput, bool) {
```
cmurphy, line 333: Is this returning a non-pointer `eks.UpdateNodegroupConfigInput` because of the tests?

package/Dockerfile
```diff
@@ -4,4 +4,5 @@ ENV KUBECONFIG /root/.kube/config
 RUN microdnf update -y && \
     rm -rf /var/cache/yum
 COPY bin/eks-operator /usr/bin/
+USER 1001
```
thedadams, line 7: This can be removed. A more extensive change is coming for the Dockerfiles.

controller/nodegroup_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/eks"
+	v1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
+	"sort"
+	"testing"
+
+
+	"github.com/stretchr/testify/assert"
```
thedadams, line 11: nit (personal opinion): I don't really like testify, especially when only used for asserts.

controller/nodegroup_test.go
```diff
@@ -0,0 +1,132 @@
+package controller
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/eks"
+	v1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
+	"sort"
+	"testing"
+
+
```
thedadams, line 10: Remove line.

controller/nodegroup_test.go
```diff
@@ -0,0 +1,132 @@
...
+				ScalingConfig: &eks.NodegroupScalingConfig{
+					MinSize: aws.Int64(1),
+					MaxSize: aws.Int64(1),
+				}},
+			expectedNgNeedsUpdate: true,
+		},
+	}
+	for _, testCase := range testCases {
+		ngUpdateInput, ngNeedsUpdate := getNodegroupConfigUpdate(testCase.clusterName, testCase.ng1, testCase.ng2)
+		if ngUpdateInput.Labels != nil && len(ngUpdateInput.Labels.RemoveLabels) > 0{
```
thedadams, line 124: nit: 
```suggestion
		if ngUpdateInput.Labels != nil && len(ngUpdateInput.Labels.RemoveLabels) > 0 {
```

utils/map.go
```diff
@@ -33,8 +33,9 @@ func GetKeysToDelete(tags map[string]string, upstreamTags map[string]string) []*
 	}
 
 	var updateUntags []*string
-	for key, val := range upstreamTags {
-		if len(tags) == 0 || tags[key] != val {
+	for key := range upstreamTags {
+		_, ok := tags[key]
+		if len(tags) == 0 || !ok {
```
thedadams, line 38: If `len(tags) == 0`, then `ok` will be `false`.
```suggestion
		if !ok {
```

thedadams (changes requested):

controller/nodegroup_test.go
rmweir: We use it all over rancher. Do you just think it's not worth bringing in for assert only? Why don't you like testify?

thedadams: I know it is used in a lot of Go packages. I haven't found that it provides me a lot of value, especially assert. Error handling in Go is very verbose, and using assert feels like it's from a different language.

controller/nodegroup.go
```diff
@@ -327,3 +327,50 @@ func getImageRootDeviceName(imageID *string, ec2Service *ec2.EC2) (*string, erro
...
+	}
+
+	if ng.MaxSize != nil {
+		nodegroupConfig.ScalingConfig.MaxSize = ng.MaxSize
+		if aws.Int64Value(upstreamNg.MaxSize) != aws.Int64Value(ng.MaxSize) {
+			sendUpdateNodegroupConfig = true
+		}
+	}
+
+	fmt.Println(nodegroupConfig.Labels)
```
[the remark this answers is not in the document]
rmweir: Removed.

controller/nodegroup.go
rmweir: I'm returning a value because of the test and I'm not returning a pointer so that the value goes on the stack and not the heap. My understanding is it's best to do this unless you have a good reason to use a pointer as not using one should lead to less bugs and slightly better performance. I'm probably not super consistent with it and it's not a huge optimization but it's something that was discussed in some go classes we took from Ardan labs.

package/Dockerfile
rmweir: I think I accidentally based off a PR of yours or something as it had you co-authored and I don't remember adding anything like that, rebased.

cmurphy (approved):

thedadams (approved):

controller/nodegroup_test.go
```diff
@@ -0,0 +1,132 @@
+package controller
```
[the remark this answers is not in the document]
cbron: +1

controller/nodegroup_test.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+func TestGetNodegroupConfigUpdate(t *testing.T) {
+	type nodegroupUpdateTestCase struct {
+		clusterName string
+		ng1 v1.NodeGroup
+		ng2 v1.NodeGroup
+		expectedNgUpdateInput eks.UpdateNodegroupConfigInput
+		expectedNgNeedsUpdate bool
+	}
```
cbron, line 20: nit: its nice to have a basic case title with each of them. Like "delete and update of old labels".

controller/nodegroup.go
cbron: I know we only broke it out for testing, but I'm always happy when a functional block gets put into a separate method. IMO we have too many run-on functions.

controller/nodegroup.go
```diff
@@ -327,3 +327,50 @@ func getImageRootDeviceName(imageID *string, ec2Service *ec2.EC2) (*string, erro
...
+			sendUpdateNodegroupConfig = true
+			nodegroupConfig.Labels = &eks.UpdateLabelsPayload{
+				RemoveLabels:      unlabels,
+				AddOrUpdateLabels: labels,
+			}
+		}
+	}
+
+	if ng.DesiredSize != nil {
+		nodegroupConfig.ScalingConfig.DesiredSize = ng.DesiredSize
```
cbron, line 355: So to be clear, the change here is this line moving out of the conditional. So that if only label updates are going out, the scalingConfigs now go out with them either way, even if they aren't changed from upstream ?

cbron (approved):
If my assumptions are correct then lgtm

controller/nodegroup.go
rmweir: Yep.

controller/nodegroup_test.go
rmweir: Added.

rmweir:
@cbron added descriptions to each test case, ran go fmt, and fixed imports block.

controller/nodegroup_test.go
cbron: Oh I actually meant in the struct, so the test output would include the title. But we can save that for next time, as long as the case summary exists alongside the test.

cbron (approved):

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/rancher/wrangler v0.7.3-0.20201020003736-e86bc912dfac
 	github.com/rancher/wrangler-api v0.6.1-0.20200427172631-a7c2f09b783e
 	github.com/sirupsen/logrus v1.4.2
+	github.com/stretchr/testify v1.4.0
```
cbron, line 14: was there not a go.sum change here ?

rmweir: Looks like it, just tried `go mod tidy` tidy on main branch and nothing.

PR #46, opened 2021-06-29
Run c_rehash before launching the operator

thedadams:
The SLE image requires that c_rehash is run before additional trusted CAs
will be detected. Therefore, an entrypoint script is added to accomplish
this.

In addition, fsGroup is added so that c_rehash (which is run as the
eks-operator user) can read the cert that is mounted.

Issue:
https://github.com/rancher/rancher/issues/31846

### DIFF
--- a/charts/eks-operator/templates/deployment.yaml
+++ b/charts/eks-operator/templates/deployment.yaml
@@ -14,6 +14,9 @@ spec:
         ke.cattle.io/operator: eks
     spec:
       serviceAccountName: eks-operator
+      securityContext:
+        fsGroup: 1007
+        runAsUser: 1007
       containers:
       - name: eks-operator
         image: {{ template "system_default_registry" . }}{{ .Values.eksOperator.image.repository }}:{{ .Values.eksOperator.image.tag }}
@@ -26,8 +29,16 @@ spec:
         - name: NO_PROXY
           value: {{ .Values.noProxy }}
 {{- if .Values.additionalTrustedCAs }}
+        # eks-operator mounts the additional CAs in two places:
         volumeMounts:
-          - mountPath: /etc/ssl/certs/ca-additional.pem
+            # This directory is owned by the eks-operator user so c_rehash works here.
+          - mountPath: /etc/rancher/ssl/ca-additional.pem
+            name: tls-ca-additional-volume
+            subPath: ca-additional.pem
+            readOnly: true
+            # This directory is root-owned so c_rehash doesn't work here,
+            # but the cert is here in case update-ca-certificates is called in the future or by the OS.
+          - mountPath: /etc/pki/trust/anchors/ca-additional.pem
             name: tls-ca-additional-volume
             subPath: ca-additional.pem
             readOnly: true
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -4,6 +4,14 @@ RUN zypper -n update && \
     rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/*
 RUN useradd --uid 1007 eks-operator
 ENV KUBECONFIG /home/eks-operator/.kube/config
+ENV SSL_CERT_DIR /etc/rancher/ssl
+
 COPY bin/eks-operator /usr/bin/
+COPY package/entrypoint.sh /usr/bin
+RUN chmod +x /usr/bin/entrypoint.sh
+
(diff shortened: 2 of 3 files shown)


cmurphy (approved):

thedadams:
I added an additional mount for the CA at `/etc/pki/trust/anchors/` in case `update-ca-certificates` is ever called in the container. This will allow `eks-operator` to run `c_rehash` on startup, as well as handle any other calls to `update-ca-certificates` to occur.

charts/eks-operator/templates/deployment.yaml
```diff
@@ -27,7 +30,11 @@ spec:
           value: {{ .Values.noProxy }}
 {{- if .Values.additionalTrustedCAs }}
         volumeMounts:
-          - mountPath: /etc/ssl/certs/ca-additional.pem
+          - mountPath: /etc/rancher/ssl/ca-additional.pem
+            name: tls-ca-additional-volume
+            subPath: ca-additional.pem
+            readOnly: true
+          - mountPath: /etc/pki/trust/anchors/ca-additional.pem
```
cmurphy, line 37: Is this meant to only mount `ca-additional.pem`, or did you want to mount the whole anchors directory?

Why not `/etc/ssl/certs`? I don't see any reference to `ca-additional.pem` being anywhere but `/etc/rancher/ssl` or `/etc/ssl/certs`

thedadams: We only want to mount the ca-additional.pem file.

The reason for the directory is here: https://documentation.suse.com/suse-caasp/4.5/html/caasp-admin/_troubleshooting_3.html#_error_x509_certificate_signed_by_unknown_authority

cmurphy (approved):

rmweir (approved):
LGTM

charts/eks-operator/templates/deployment.yaml
```diff
@@ -27,7 +30,11 @@ spec:
           value: {{ .Values.noProxy }}
 {{- if .Values.additionalTrustedCAs }}
         volumeMounts:
-          - mountPath: /etc/ssl/certs/ca-additional.pem
+          - mountPath: /etc/rancher/ssl/ca-additional.pem
```
cbron, line 33: In the Rancher PR you only added it to `/etc/pki/anchors....` and removed the custom Rancher dir. Why does eks still need the custom rancher dir ?

package/Dockerfile
```diff
@@ -4,6 +4,14 @@ RUN zypper -n update && \
     rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/*
 RUN useradd --uid 1007 eks-operator
 ENV KUBECONFIG /home/eks-operator/.kube/config
+ENV SSL_CERT_DIR /etc/rancher/ssl
```
cbron, line 7: Didn't realize before this PR that you could configure openssl this way.

package/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+if [ -x "$(command -v c_rehash)" ]; then
+  c_rehash
```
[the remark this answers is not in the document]
thedadams: `update-ca-certificates` requires root. `c_rehash` can be run as long as the eks-operator user owns the directory.

charts/eks-operator/templates/deployment.yaml
thedadams: Because the eks-operator user owns the `/etc/rancher/ssl` directory so `c_rehash` works there.

package/entrypoint.sh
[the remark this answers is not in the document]
cbron: Can you add this as a comment in the file

charts/eks-operator/templates/deployment.yaml
cbron: Needs a comment in this file saying _"as opposed to rancher which runs `update-ca-certificates`, we run c_rehash and need both dirs...."_

thedadams: Done.

package/entrypoint.sh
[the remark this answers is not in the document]
thedadams: Done.

cbron (approved):

PR #51, opened 2021-07-15
Update AWS SDK

sgran:
This removes validation for nodegroup min and desired size to be greater
than zero, allowing cluster autoscaler to scale from 0.

Signed-off-by: Stephen Gran <steve.gran@anaplan.com>


furkatgofurov7:
This PR is redundant and needs a rebase since we are already using aws SDK version v1.45.6, closing it.

PR #60, opened 2022-02-16
Add security groups to EKS launch template

annablender:
Address https://github.com/rancher/rancher/issues/34749

The `eks-operator` was adding [security groups](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html) when creating a cluster, but not when creating a launch template. When creating an EKS cluster in Rancher, if you specify an additional security group, it will now show up in the both the launch template and node instances.


controller/nodegroup.go
```diff
@@ -115,6 +123,7 @@ func buildLaunchTemplateData(group eksv1.NodeGroup, ec2Service *ec2.EC2) (*ec2.R
 			},
 		},
 		TagSpecifications: utils.CreateTagSpecs(group.ResourceTags),
+		SecurityGroupIds: securityGroupIds,
```
thedadams, line 126: Instead of creating a new slice manually, you can use `aws.StringSlice(securityGroups)` here directly.

controller/nodegroup.go
```diff
@@ -102,6 +102,14 @@ func buildLaunchTemplateData(group eksv1.NodeGroup, ec2Service *ec2.EC2) (*ec2.R
 		}
 	}
 
+	// For a non-default VPC, you must use security group IDs. You cannot specify both a security group ID and security
+	// name in the same request.
+	// So we grab the security group IDs to add to the launch template.
```
thedadams, line 107: This comment is a little confusing because this code doesn't reference any "security name." I think adding the security group IDs to the launch template is enough context as to what is happening here.

thedadams (changes requested):

thedadams (dismissed):

thedadams (approved):

jiaqiluo (approved):
LGTM 💯

PR #61, opened 2022-02-17
Changes required to support mixed workload clusters.

phillipsj:
This adds new annotations for app and marketplace UI changes. It also adding the correct tolerations and selectors for running in a mixed workload cluster.

* https://github.com/rancher/rancher/issues/36401

Verified that it will deploy in a mixed workload cluster successfully after the changes.

eks-operator-with-changes


thedadams (approved):

rosskirkpat (approved):
LGTM

PR #65, opened 2022-05-19
Enable node group node role

Sheldons92:
As per request, bringing the code from https://github.com/rancher/eks-operator/pull/53 up to date with master.
For https://github.com/rancher/rancher/issues/29071


controller/nodegroup.go
```diff
@@ -261,13 +261,17 @@ func createNodeGroup(config *eksv1.EKSClusterConfig, group eksv1.NodeGroup, eksS
...
-	output, err := createStack(svc, fmt.Sprintf("%s-node-instance-role", config.Spec.DisplayName), config.Spec.DisplayName, finalTemplate, []string{cloudformation.CapabilityCapabilityIam}, []*cloudformation.Parameter{})
-	if err != nil {
-		return "", err
+	if aws.StringValue(group.NodeRole) == "" {
+		finalTemplate := fmt.Sprintf(templates.NodeInstanceRoleTemplate, getEC2ServiceEndpoint(config.Spec.Region))
+		output, err := createStack(svc, fmt.Sprintf("%s-node-instance-role", config.Spec.DisplayName), config.Spec.DisplayName, finalTemplate, []string{cloudformation.CapabilityCapabilityIam}, []*cloudformation.Parameter{})
+		if err != nil {
+			return "", err
+		}
+		nodeGroupCreateInput.NodeRole = aws.String(getParameterValueFromOutput("NodeInstanceRole", output.Stacks[0].Outputs))
```
thedadams, line 270: The name of this node role will be something complicated like `donnie-test-node-instance-role-NodeInstanceRole-XXW9GQYTQWIZ`. It would be a good idea to put this on the status of the `EKSClusterConfig` object, so it can be tracked in other places.

thedadams (changes requested):
In addition the comments on the code, this PR needs to be rebased. The convention is to have one commit with the changes you are making and another commit with the changes after running `go generate`.

thedadams:
@Sheldons92 Thank you for rebasing. However, your PR should have two commits: one with your explicit changes and one with the changes of `go generate`.

Additionally, you are assuming that the node role names are too simple. The generated one has a much more complicated name than you looking for. Please see me comments above.

thedadams:
It was my intention to get this done today, but it seems that I am not able to because I am having a strange issue running `go generate` locally.

[This](https://github.com/thedadams/eks-operator/tree/enable-node-group-node-role) branch has the changes in it. There might be a few issues related to types or something, but for the most part it should work after `go generate` is run.

In addition, changes need to be made in rancher/rancher:
1. A `GeneratedNodeRole` field should be added to the EKSStatus of the v3 cluster object and this should be updated from the `eksclusterconfig` object in the same way that the `ManagedLaunchTemplateID` is. See here for how the `ManagedLaunchTemplateID` is pulled of the `eksclusterconfig` object and put on the `EKSStatus` field.
2. The `GeneratedNodeRole` needs to be passed to the `BuildUpstreamClusterSpec` call [here](https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go#L83). This is to ensure we don't accidentally sync the node role to the cluster spec by not recognizing it is the one generated by Rancher.

mjura:
It was already fixed

PR #66, opened 2022-06-17
Remove metadata from EKS errors to stop log spamming

annablender:
Issue: https://github.com/rancher/rancher/issues/37650 

 
## Problem
 

When a call to `UpdateNodeGroupVersion` fails due to a custom AMI version causing node provisioning to fail in EKS, the failure is expected but lasso spams the EKS logs every second. This happens because the error message that is returned from EKS has metadata with a unique `requestID` in the body. Since the error message is different every time, [this code](https://github.com/rancher/eks-operator/blob/1b8dfe2ea04eee43fd362c6da0e1cef2a09cb600/controller/eks-cluster-config-handler.go#L122) is skipped and the [Status gets updated](https://github.com/rancher/eks-operator/blob/1b8dfe2ea04eee43fd362c6da0e1cef2a09cb600/controller/eks-cluster-config-handler.go#L133) every time which reenqueues the handler immediately instead of doing a backoff reenqueue and is spamming the logs.

## Solution
 

My solution is to remove the metadata from the EKS logs

What happens now when an EKS error is logged?
- Raw err message is logged
- Remove metadata from the error. If removing metadata fails because the EKS error is not in the expected format, continue with the original logic to prevent a regression.  The new error string will contain the cluster name, error message, and node group.
- If the Status doesn't already contain the error (without metadata), add the error to the Status.

image

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v15 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/util/retry"
 )
 
@@ -109,13 +110,19 @@ func (h *Handler) recordError(onChange func(key string, config *eksv1.EKSCluster
 		}
 		if err != nil {
 			if !strings.Contains(err.Error(), "currently has update") {
-				// the update is valid in that the controller should retry but there is
-				// no actionable resolution as far as a user is concerned. An update
-				// that has either been initiated by eks-operator or another source is
-				// already in progress. It is possible an update is not being immediately
-				// reflected in upstream cluster state. The config object will reenter the
-				// controller and then the controller will wait for the update to finish.
+				// The update is valid in that the controller should retry but there is no actionable resolution as far
+				// as a user is concerned. An update has either been initiated by the eks-operator or another source
+				// is already in progress. It is possible an update is not being immediately reflected in the upstream
+				// cluster state. The config object will reenter the controller and then the controller will wait for
+				// the update to finish.
 				message = err.Error()
+			}
+
+			messageNoMeta, err := removeErrorMetadata(message)
+			if err != nil {
+				logrus.Errorf("Error removing metadata from failure message: %s", err.Error())
+			} else {
+				message = messageNoMeta
 			}
 		}
 
@@ -139,6 +146,54 @@ func (h *Handler) recordError(onChange func(key string, config *eksv1.EKSCluster
 		}
 		return config, err
 	}
+}
+
+func removeErrorMetadata(message string) (string, error) {
+
+	// failure message
+	type RespMetadata struct {
+		StatusCode int    `json:"statusCode"`
+		RequestID  string `json:"requestID"`
+	}
+
+	type Message struct {
+		RespMetadata  RespMetadata `json:"respMetadata"`
+		ClusterName   string       `json:"clusterName"`
+		Message_      string       `json:"message_"`
+		NodegroupName string       `json:"nodegroupName"`
+	}
+
+	// failure message with no meta
+	type FailureMessage struct {
+		ClusterName   string `json:"clusterName"`
+		Message_      string `json:"message_"`
+		NodegroupName string `json:"nodegroupName"`
+	}
+
+	// Remove the first line of the message because it usually contains the name of an Amazon EKS error type that
+	// implements Seralizable (ex: ResourceInUseException). That name is unpredictable depending on the error. We
+	// only need cluster name, message, and node group.
+	index := strings.Index(message, "{"); if index == -1 {
+		return "", fmt.Errorf("message body not formatted as expected")
+	}
+	message = message[index:]
+
+	// unmarshal json error to an object
+	in := []byte(message)
+	failureMessage := Message{}
+	err := yaml.Unmarshal(in, &failureMessage); if err != nil {
+		return "", err
+	}
+
+	// add error message fields without metadata to new object
+	failureMessageNoMeta := FailureMessage{
+		ClusterName: failureMessage.ClusterName,
+		Message_: failureMessage.Message_,
+		NodegroupName: failureMessage.NodegroupName,
+	}
+
+	str := fmt.Sprintf("%#v", failureMessageNoMeta)
+	return str, nil
 }
 
 func (h *Handler) OnEksConfigRemoved(_ string, config *eksv1.EKSClusterConfig) (*eksv1.EKSClusterConfig, error) {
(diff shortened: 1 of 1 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -109,13 +110,16 @@ func (h *Handler) recordError(onChange func(key string, config *eksv1.EKSCluster
...
+				// The update is valid in that the controller should retry but there is no actionable resolution as far
+				// as a user is concerned. An update that has neither been initiated by the eks-operator or another
+				// source is in already in progress. It is possible an update is not being immediately reflected in
+				// upstream cluster state. The config object will reenter the controller and then the controller will
+				// wait for the update to finish.
 				message = err.Error()
+
+				// If we have hit a retriable error, we don't know whether that error will resolve or not. Wait in this
+				// case to avoid spamming the EKS logs and thrashing other updates.
+				time.Sleep(eksRecordErrorTimeout * time.Second)
```
cmurphy, line 122: All the controllers have an `EnqueueAfter` method which allows you to reenqueue the handler after a time period:

https://github.com/rancher/eks-operator/blob/1b8dfe2ea04eee43fd362c6da0e1cef2a09cb600/controller/eks-cluster-config-handler.go#L61
https://github.com/rancher/eks-operator/blob/1b8dfe2ea04eee43fd362c6da0e1cef2a09cb600/controller/eks-cluster-config-handler.go#L256

Is there a way we could use that instead? Using a sleep here will just make the handler block waiting for the time to finish instead of returning control to the controller.

annablender: Using `eksEnqueueAfter` still spams the logs every second. Using `Sleep` prevents the log spamming from what I can see

jakefhyde: > LGTM, I also wonder if there is a better approach for solving this issue. Looking forward to hearing others' feedback.

I solved this at one point in encryption key rotation by setting an annotation with the time of the last sleep, and checking if the required interval (usually 1 second shorter than what we `EnqueueAfter` with) had passed since, and if it hadn't to not enqueue the object and return. There's also some fun stuff you can do with channels and tickers, but it's sort of niche and hard to get right.

jakefhyde:
> LGTM, I also wonder if there is a better approach for solving this issue. Looking forward to hearing others' feedback.

I solved this at one point in encryption key rotation by setting an annotation with the time of the last sleep, and checking if the required interval (usually 1 second shorter than what we `EnqueueAfter` with) had passed since, and if it hadn't to not enqueue the object and return. There's also some fun stuff you can do with channels and tickers, but it's sort of niche and hard to get right.

controller/eks-cluster-config-handler.go
```diff
@@ -109,13 +109,17 @@ func (h *Handler) recordError(onChange func(key string, config *eksv1.EKSCluster
...
+				// as a user is concerned. An update that has neither been initiated by the eks-operator or another
+				// source is in already in progress. It is possible an update is not being immediately reflected in
+				// upstream cluster state. The config object will reenter the controller and then the controller will
+				// wait for the update to finish.
 				message = err.Error()
+
+				// If we have hit a retriable error, we don't know whether that error will resolve or not. Wait in this
+				// case to avoid spamming the EKS logs and thrashing other updates.
+				logrus.Infof("waiting for cluster [%s] update to complete", config.Name)
+				time.Sleep(30 * time.Second)
```
rmweir, line 122: I don't think we want to block the thread like this. I think it's actually returning an error which reenqueues the controller since the status is a subresource iirc, so changes to it do cause a reenqueue. Reenqueueing on errors is the backboen to the framework and shouldn't be skipped like this. I would also think that we want the first occurrence to log immediately. There is a builtin backoff that should slow it down over time iirc. If you can show us an example of the offending error and we may be able to think of a different solution.

rmweir (changes requested):

controller/eks-cluster-config-handler.go
```diff
@@ -108,15 +109,22 @@ func (h *Handler) recordError(onChange func(key string, config *eksv1.EKSCluster
...
+				// wait for the update to finish.
 				message = err.Error()
 			}
+
+			messageNoMeta, err := removeErrorMetadata(message)
+			if err != nil {
+				logrus.Errorf("Error removing metadata from failure message: %s", err.Error())
+			} else {
+				message = messageNoMeta
+			}
```
jiaqiluo, line 127: nit: simplify the logic 
```suggestion
			} 
			message = messageNoMeta
```

jiaqiluo (approved):

controller/eks-cluster-config-handler.go
annablender: I believe my logic is correct here because we don't want the message to be set to `messageNoMeta` if the metadata removal returned an err

jiaqiluo: I see. yeah, it makes sense. :)

jiaqiluo (approved):

controller/eks-cluster-config-handler.go
```diff
@@ -141,6 +148,52 @@ func (h *Handler) recordError(onChange func(key string, config *eksv1.EKSCluster
...
+		ClusterName   string `json:"clusterName"`
+		Message_      string `json:"message_"`
+		NodegroupName string `json:"nodegroupName"`
+	}
+
+	// Remove the first line of the message because it usually contains the name of an Amazon EKS error type that
+	// implements Seralizable (ex: ResourceInUseException). That name is unpredictable depending on the error. We
+	// only need cluster name, message, and node group.
+	index := strings.Index(message, "{")
+	message = message[index:]
```
HarrisonWAffel, line 177: `Strings.Index` will return `-1` if the substring isn't found, in which case `message[index:]` would panic. Will all error messages pass through this function?

rmweir: Should check for -1 and return error if true.

rmweir (changes requested):

controller/eks-cluster-config-handler.go
```diff
@@ -109,14 +110,20 @@ func (h *Handler) recordError(onChange func(key string, config *eksv1.EKSCluster
...
-				// no actionable resolution as far as a user is concerned. An update
-				// that has either been initiated by eks-operator or another source is
-				// already in progress. It is possible an update is not being immediately
-				// reflected in upstream cluster state. The config object will reenter the
-				// controller and then the controller will wait for the update to finish.
+				// The update is valid in that the controller should retry but there is no actionable resolution as far
+				// as a user is concerned. An update that has neither been initiated by the eks-operator or another
+				// source is in already in progress. It is possible an update is not being immediately reflected in
+				// upstream cluster state. The config object will reenter the controller and then the controller will
+				// wait for the update to finish.
```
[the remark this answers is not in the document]
annablender: Formatting. It looked really chunky on the page so I lengthened the comment to make it more readable and match the other lines

rmweir (dismissed):
LGTM

rmweir (approved):

HarrisonWAffel (dismissed):

jiaqiluo (approved):
Please squash two commits into one before merging. the 1st commit message is misleading now.

jiaqiluo (approved):

HarrisonWAffel (approved):

PR #68, opened 2022-07-06
Add cluster security group to node launch template

annablender:
Issue: https://github.com/rancher/rancher/issues/38014 

 
## Problem
 

This issue is a regression when an EKS cluster is provisioned via the rancher UI, the [default security group](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/default-custom-security-groups.html#default-security-group) is not appended to the node group when a launch template is not specified.
 
## Solution
 

My solution here is to create a new launch template version and apply it to the EC2 nodes after the cluster is finished creating. EKS does not generate the default security group until after the cluster finishes creating. Specifically in `eks operator`, we now add a bool `addDefaultSecurityGroup` to mark that it needs to be added to the node group, update the `config.Status` with the vpc default security group included, create a new launch template version and apply it to the nodes.

Functionality notes

The security group update takes some time, so when you provision an EKS cluster with no launch template specified, the update will take an extra 5-10 mins on top of provisioning. I usually see 2 nodes clusters become fully active in Rancher after about 25 mins. That time bump is due to EKS creating a new launch template, deleting the old one, and recycling the nodes in the node group to use the new launch template.

Sometimes, the cloud provider is slow to update the cluster and `config.Status.SecurityGroups` doesn't get updated until after a few syncs of the operator. This results in 2-3 new launch template versions getting created before the update completes. The old ones are immediately cleaned up by rancher, so @rmweir have agreed this is fine as long as EKS is not getting overloaded with template objects.
 
## Testing
 

1. Provision a 2 node EKS cluster via the rancher UI with 1-3 custom security groups and NO launch template
2. Provision a 2 node EKS cluster via the rancher UI with 1-3 custom security groups and specified launch template

Verify that both clusters become active.


controller/eks-cluster-config-handler.go
```diff
@@ -1185,14 +1189,25 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
 			lt := ng.LaunchTemplate
 
+			// once the cluster is created, update the node group launch template with the default cluster security
+			// group generated by EKS
+			addDefaultClusterSecurityGroup := false
+			if !reflect.DeepEqual(config.Status.SecurityGroups, upstreamSpec.SecurityGroups) {
+				logrus.Infof("adding default security group to node group [%s] for cluster [%s]", aws.StringValue(upstreamNg.NodegroupName), config.Name)
+				addDefaultClusterSecurityGroup = true
+				config = config.DeepCopy()
+				config.Status.SecurityGroups = upstreamSpec.SecurityGroups
```
rmweir, line 1205: WHy does this need to be done for every upstream nodegroup?

controller/eks-cluster-config-handler.go
```diff
@@ -1185,14 +1189,25 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
+		// there exists a launch template for this node group
 		if upstreamNg.LaunchTemplate != nil {
 			upstreamTemplateVersion := aws.Int64Value(upstreamNg.LaunchTemplate.Version)
 			var err error
 			lt := ng.LaunchTemplate
 
+			// once the cluster is created, update the node group launch template with the default cluster security
+			// group generated by EKS
+			addDefaultClusterSecurityGroup := false
+			if !reflect.DeepEqual(config.Status.SecurityGroups, upstreamSpec.SecurityGroups) {
```
rmweir, line 1201: Would it be possible for these to have equal elements but in different orders?

rmweir (changes requested):

controller/eks-cluster-config-handler.go
annablender: Because every node instance in all the node groups for the cluster need to have the AWS cluster default security group attached, since it's cluster level

controller/eks-cluster-config-handler.go
annablender: Not in the case of a provisioned EKS cluster. Once rancher defined security groups are set, they are pulled in that same order from upstream and the cluster default security group is always appended to the _end_ of that list. It will always have the order [<rancher sgs (in order selected in the rancher UI>, <default sg>]. Once user defined security groups are selected, they cannot be modified

controller/eks-cluster-config-handler.go
rmweir: but upstreamSpec isn't being updated so isn't one write of this enough? Like how would `!reflect.DeepEqual(config.Status.SecurityGroups, upstreamSpec.SecurityGroups)` ever be met after the first iteration?

rmweir (changes requested):

rosskirkpat (approved):

controller/eks-cluster-config-handler.go
```diff
@@ -1171,6 +1175,16 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
 	}
 
+	// once the cluster is created, update the node group launch template with the default security group
+	// generated by EKS
+	addDefaultSecurityGroup := false
+	if !reflect.DeepEqual(config.Status.SecurityGroups, upstreamSpec.SecurityGroups) {
+		logrus.Infof("adding default security group to node groups for cluster [%s]", config.Name)
+		addDefaultSecurityGroup = true
+		config = config.DeepCopy()
+		config.Status.SecurityGroups = upstreamSpec.SecurityGroups
```
rmweir, line 1185: I think whenever we do a change to config we want to update it with the client because we don't want to end up returning it from the handler if it's not the true state of the object and the client will give us the true state whether there's an error or not. See here for example: https://github.com/rancher/eks-operator/pull/68/files#diff-a74e30a65934de22a0dafc6fd3a31ff5c5c81d47e8fbd43f4f740c419977b565R1125.

rmweir (changes requested):

jakefhyde (approved):

rmweir (approved):
LGTM

PR #73, opened 2022-09-13
[v2.7.1] Enable support for custom node roles for node groups

annablender:
Issue: https://github.com/rancher/rancher/issues/29071

 
## Problem

 
Enable custom node roles for node groups.

## Solution

Changes
* NodeRole (string) is added to the NodeGroup struct for a user-defined node group node role in Rancher
* GeneratedNodeRole (string) is added to the config.Spec.EKSStatus for a node group node role generated and tracked by Rancher if a user doesn't provide one.

## Testing

1. Provision 2 node EKS cluster with no specified node group node role. Cluster should come up active and node role should be viewable on spec.config.EKSStatus in the v3 object
2. Provision 2 node EKS cluster with user-defined node group node role
3. Provision 2 node EKS cluster with 2 node groups, 1 with no specified node group node role and 1 with a user-defined node-group role.

In the [UI issue](https://github.com/rancher/dashboard/issues/4568), it is specified

```
The value passed to the Rancher API should be a string. If no role is selected for a nodegroup, then the empty string should be passed.
```

I simulated rancher UI input by setting the empty string and a valid user-defined string on the existing node group `NodeRole` field in the code and verify the expected flow happens in the operator. `NodeRole` cannot be nil. _Changes have been tested and verified with the PR https://github.com/rancher/rancher/pull/38939._


jakefhyde (approved):

rmweir (changes requested):
Are you using a test catalog in your setup that has the new crd?

controller/nodegroup.go
```diff
@@ -253,13 +253,27 @@ func createNodeGroup(config *eksv1.EKSClusterConfig, group eksv1.NodeGroup, eksS
...
+			output, err := createStack(svc, fmt.Sprintf("%s-node-instance-role", config.Spec.DisplayName), config.Spec.DisplayName, finalTemplate, []string{cloudformation.CapabilityCapabilityIam}, []*cloudformation.Parameter{})
+			if err != nil {
+				// If there was an error creating the node role stack, return an empty launch template
+				// version and the error.
+				return "", err
+			}
+
+			// Ensure that the change persists for the calling function. No need to save here because
+			// the config object will get saved by the calling function.
+			*config = *config.DeepCopy()
```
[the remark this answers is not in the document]
annablender: It was experimental to persist the change of `Status.GeneratedNodeRole = <value>` back to the calling function, but it's not working from what I've tested so I updated the fix to return the generated node role to the caller function. The same thing is actually being done with other fields set on the status (managedlaunchtemplateId, managedLaunchTemplateVersions) so it follows design. I still have to test it with the set of updates from rancher/rancher

annablender:
@rmweir Are you using a test catalog in your setup that has the new crd? -> I'm testing a local version of the eks operator on rainbow 2.7, and also testing both the rancher and operator changes together using a local docker image of the eks operator repo, thank you @jakefhyde

controller/nodegroup.go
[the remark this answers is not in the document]
annablender: Ok, so I was testing without an updated eks crd.

Once I fixed that, the deepCopy here does work. It persists the change to `config.Status` back to the calling function instead of the change being lost. I realized it's also easier to set the status in nodegroup.go, because we are checking the value of `GeneratedNodeRole` here. I've updated the fix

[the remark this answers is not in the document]
jakefhyde: I don't think the issue is whether or not the config object gets copied, but rather the `DeepCopy` call doesn't actually do anything useful. As @rmweir pointed out, `DeepCopy` is used when we don't want to modify the original data being pointed to, in this case we just do an in place object copy so we aren't actually doing anything useful.

controller/eks-cluster-config-handler.go
```diff
@@ -807,7 +810,7 @@ func (h *Handler) waitForCreationComplete(config *eksv1.EKSClusterConfig, eksSer
 }
 
 // buildUpstreamClusterState
-func BuildUpstreamClusterState(name, managedTemplateID string, clusterState *eks.DescribeClusterOutput, nodeGroupStates []*eks.DescribeNodegroupOutput, ec2Service *ec2.EC2, includeManagedLaunchTemplate bool) (*eksv1.EKSClusterConfigSpec, string, error) {
+func BuildUpstreamClusterState(name, managedTemplateID string, generatedNodeRoleID string, clusterState *eks.DescribeClusterOutput, nodeGroupStates []*eks.DescribeNodegroupOutput, ec2Service *ec2.EC2, includeManagedLaunchTemplate bool) (*eksv1.EKSClusterConfigSpec, string, error) {
```
HarrisonWAffel, line 813: `generatedNodeRoleID` was added to the function signature, but I don't see it being used anywhere. Is it still needed?

controller/nodegroup.go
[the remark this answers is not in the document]
rmweir: Again, I don't think this does anything and believe it should be removed if there is no justification for keeping it.

controller/eks-cluster-config-handler.go
HarrisonWAffel: I think other than this, and the deep-copy conversation above, this lgtm

controller/nodegroup.go
```diff
@@ -253,20 +253,35 @@ func createNodeGroup(config *eksv1.EKSClusterConfig, group eksv1.NodeGroup, eksS
...
+			if err != nil {
+				// If there was an error creating the node role stack, return an empty launch template
+				// version and the error.
+				return "", err
+			}
+
+			// Ensure the change persists for the calling function. No need to save here because the
+			// config object will get saved by the calling function.
+			*config = *config.DeepCopy()
+			config.Status.GeneratedNodeRole = getParameterValueFromOutput("NodeInstanceRole", output.Stacks[0].Outputs)
```
rmweir, line 269: Where does this get saved? I think it should probably be added here: https://github.com/rancher/eks-operator/pull/73/files#diff-a74e30a65934de22a0dafc6fd3a31ff5c5c81d47e8fbd43f4f740c419977b565R1155, and this function should probably not write to the config at all as it is a side effect of what it's main goal is, creating the nodegroup in eks.

go.mod
```diff
@@ -12,9 +12,12 @@ require (
...
-	golang.org/x/sys v0.0.0-20220715151400-c0bba94af5f8 // indirect
 	k8s.io/api v0.21.2
 	k8s.io/apimachinery v0.21.2
 	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
 )
+
+require (
+	golang.org/x/net v0.0.0-20220708220712-1185a9018129 // indirect
+	golang.org/x/sys v0.0.0-20220715151400-c0bba94af5f8 // indirect
+)
```
rmweir, line 23: Is there a reason why this was moved to a different block?

rmweir (changes requested):

go.mod
jakefhyde: If I had to hazard a guess, `go mod tidy` was run with `>= go1.17`

jakefhyde (changes requested):
Couple nits

go.mod
annablender: I ran it with go1.19.2 which is specified in the `go.mod.` Is this incorrect?

controller/eks-cluster-config-handler.go
annablender: This is an interesting question. Technically no, `generatedNodeRoleID` is not needed. It works how it is now. But I realize it would be best practice for rancher to pass that field in from the Status since it's a Rancher-generated field. Pulling it from EKS implies its EKS-generated. This is really only something devs care about because either way the functionality will be the same.

HarrisonWAffel: I guess I'm asking why we need to pull it in at all if we don't use it. Is there a valid reason why we may want it in the future and why adding it now would be beneficial?

annablender:
I'm retesting this issue with addressed PR comments. I'm having issues getting the `generatedNodeRole` to get fully updated on the Status when a new node group is created now. I think there's an object reference problem in the operator, but still working on it. DNM until 2.7.1 branch is out.

annablender:
This has been successfully retested on rancher `v2.7 head` with PR comments addressed.

The issue with `generatedNodeRole` was that I had accidentally mod some previous code while I was testing the issue that caused the config obj to be shadowed. So, config was being modified before a new node group was created causing k8s to throw an obj modification err when I tried to update the Status. This has been fixed and is working fine now, along with the other changes made to the issue

* `generatedNodeRole` is now set on the Status in the handler instead of in nodegroup.go because it doesn't coincide with the utility of that function
* unnecessary deepCopies have been removed

controller/eks-cluster-config-handler.go
```diff
@@ -470,6 +470,12 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
 }
 
 func (h *Handler) create(config *eksv1.EKSClusterConfig, sess *session.Session, eksService *eks.EKS) (*eksv1.EKSClusterConfig, error) {
+	// test
+	*config = *config.DeepCopy()
+	nodeRole := ""
+	config.Spec.NodeGroups[0].NodeRole = &nodeRole
+	config.Spec.NodeGroups[1].NodeRole = &nodeRole
```
jakefhyde, line 477: looks like there is still some test code here

controller/eks-cluster-config-handler.go
```diff
@@ -1093,8 +1103,20 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
+	// Deep copy the config object here, so it's not copied multiple times for each
+	// nodegroup create/delete.
+	config = config.DeepCopy()
+
+	// test
+	if len(config.Spec.NodeGroups) > 1 {
+		config.Spec.NodeGroups[0].NodeRole = new(string)
+		str := "arn:aws:iam::270074865685:role/ablender-2nodegroups-node-instanc-NodeInstanceRole-SBF895G87LLA"
+		config.Spec.NodeGroups[1].NodeRole = &str
+	}
```
jakefhyde, line 1115: test code

jakefhyde (changes requested):

jakefhyde (approved):

rmweir (changes requested):

controller/eks-cluster-config-handler.go
```diff
@@ -1115,15 +1124,19 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
 			if err != nil {
 				return config, err
 			}
 		}
-		ltVersion, err := createNodeGroup(config, ng, eksService, ec2Service, svc)
+		ltVersion, role, err := createNodeGroup(config, ng, eksService, ec2Service, svc)
+
+		// if the generated node role hasn't been set yet and it just got generated, save
+		// it to set on the Status
+		if generatedNodeRole == "" && role != "" {
```
jakefhyde, line 1137: Just to confirm, the intention here is that the last node in the node group becomes to `config.Status.GeneratedNodeRole`? Would there be any significance to 2 node groups having different generated node roles?

rmweir (dismissed):
LGTM

jakefhyde (approved):

controller/nodegroup.go
```diff
@@ -253,21 +253,34 @@ func createNodeGroup(config *eksv1.EKSClusterConfig, group eksv1.NodeGroup, eksS
...
 	}
 
-	finalTemplate := fmt.Sprintf(templates.NodeInstanceRoleTemplate, getEC2ServiceEndpoint(config.Spec.Region))
-	output, err := createStack(svc, fmt.Sprintf("%s-node-instance-role", config.Spec.DisplayName), config.Spec.DisplayName, finalTemplate, []string{cloudformation.CapabilityCapabilityIam}, []*cloudformation.Parameter{})
-	if err != nil {
-		return "", err
+	generatedNodeRole := config.Status.GeneratedNodeRole
+
+	if aws.StringValue(group.NodeRole) == "" {
+		if config.Status.GeneratedNodeRole == "" {
```
jakefhyde, line 259: `config.Status.GeneratedRole` is not set until after the first reconciliation, so every node group will end up creating a stack. Is that by design?

controller/eks-cluster-config-handler.go
jakefhyde: Left a clarifying comment here: https://github.com/rancher/eks-operator/pull/73/files/929ccff9443e3aa0f2e3468e71e6b5b7474759ce#r1025781996

controller/eks-cluster-config-handler.go
```diff
@@ -1115,15 +1123,19 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
 			if err != nil {
 				return config, err
 			}
 		}
-		ltVersion, err := createNodeGroup(config, ng, eksService, ec2Service, svc)
+		ltVersion, generatedNodeRole, err := createNodeGroup(config, ng, eksService, ec2Service, svc)
+
+		// if a generated node role has not been set on the Status yet and it
+		// was just generated, set it
+		if config.Status.GeneratedNodeRole == "" && generatedNodeRole != "" {
```
jakefhyde, line 1136: LGTM, the only potential concerns I might have are that if an error returns later on, and we have to re-reconcile this object, it could potentially create more node groups than necessary (i.e. if another error happens between now and the time we end up updating the config object). I consider that outside of the scope of this PR because it looks like the entirety of this operator (and likely all the others) is written in a way that is contradictory to the design of wrangler's controller framework. We should revisit this at some point to prevent unnecessary thrashing by refactoring this. For example, once `config.Status.GeneratedNodeRole` is set, we should update the config object, and if it fails delete the node group and reenqueue the config object, and if it succeeds, return early since the next reconciliation will shore up the rest of the status updates. Once again, out of the scope of this PR though.

jakefhyde (approved):

rmweir (approved):
LGTM

PR #79, opened 2023-01-10
Move aws services to a separate package

alexander-demicev:
Move AWS services to a separate package and generate mocks for them.


.golangci.yml
```diff
@@ -31,11 +31,16 @@ linters-settings:
       - name: range
       - name: receiver-naming
       - name: time-naming
-      - name: unexported-return
```
richardcase, line 34: What do you think to capturing the reasons for disabling linters? Perhaps as a ADR or doc somewhere in the repo.

pkg/eks/services/cloudformation.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+type CloudFormationServiceInterface interface {
+	DescribeStacks(input *cloudformation.DescribeStacksInput) (*cloudformation.DescribeStacksOutput, error)
+	DeleteStack(input *cloudformation.DeleteStackInput) (*cloudformation.DeleteStackOutput, error)
+	CreateStack(input *cloudformation.CreateStackInput) (*cloudformation.CreateStackOutput, error)
+	DescribeStackEvents(input *cloudformation.DescribeStackEventsInput) (*cloudformation.DescribeStackEventsOutput, error)
+}
+
+type cloudFormationService struct {
```
richardcase, line 15: Do we want to check that `cloudFomrationService` implements the interface?

By doing something like one of these:

- `var _ CloudFormationServiceInterface = &cloudFormationService{}`
- `func NewCloudFormationService(sess *session.Session) CloudFormationServiceInterface`

richardcase: Same question applies to the other services....i won't add a comment for each.

richardcase (commented):
Looks good to me

.golangci.yml
alexander-demicev: actually, we can use this linter. I removed it because it didn't allow to return private structures in public functions, now I replaced structure with the interface.

richardcase: Nice

pkg/eks/services/eks.go
```diff
@@ -0,0 +1,89 @@
+package services
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/eks"
+)
+
+type EKSServiceInterface interface {
```
richardcase, line 8: Any reason we can't use the interfaces already defined for these services? For example, this is the EKS interface: https://github.com/aws/aws-sdk-go/blob/main/service/eks/eksiface/interface.go

One reason I can think of is that if we upgrade to the AWS Go SDK v2 it doesn't have these interface files.

richardcase (commented):
LGTM.....just the one question around the interfaces.

pkg/eks/services/eks.go
alexander-demicev: I'd like us to have more control here and be sure that the interfaces we use are always compatible with AWS SDK.

mjura (approved):
LGTM

richardcase (approved):
lgtm

alexander-demicev:
can you approve it again? I did a rebase after dependency update

richardcase (approved):

alexander-demicev:
once again please :D I forgot to run `go mod tidy`

richardcase (approved):

PR #80, opened 2023-01-16
Change to BCI and bump

superseb:
- Align to standardized BCI
- 15.3 is EOL


alexander-demicev (approved):

superseb:
@alexander-demicev Is it planned to create a new RC to include this in v2.7.2 or will it go in after?

PR #101, opened 2023-02-08
Unit tests for cluster creation

alexander-demicev:
This PR adds unit tests for cluster creation and github workflows

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -24,4 +24,3 @@ jobs:
     - name: Run tests
       run: |
         make test
-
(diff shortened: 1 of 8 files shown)


annablender (changes requested):
Nice job on the refactor @alexander-demicev. Just curious, but did cluster creation have to be moved to a separate folder pkg/eks to test it? Could you leave it in the original folder and specify to test just that file?

If yes, then the rest of the refactor looks fine. Make sure you test both creating and updating a field in an eks cluster before you merge this.

alexander-demicev:
@a-blender The reason for moving this in a separate package is for all hosted providers to have a similar project structure and it's easier to read smaller files. Currently, there are about 1.5k loc in the controller package file.

richardcase (approved):

PR #128, opened 2023-04-04
Fix cluster cleanup in e2e

alexander-demicev:


test/e2e/suite_test.go
```diff
@@ -301,19 +301,21 @@ var _ = AfterSuite(func() {
 	By("Cleaning up Rancher Clusters")
 
 	for _, rancherCluster := range rancherClusterList.Items {
+		if rancherCluster.Name == "local" {
```
furkatgofurov7, line 304: Not yet familiar with the repo but just for my understanding: "local" cluster is the one we never run in e2e tests but hardcoded for local usage?

furkatgofurov7 (commented):
LGTM

test/e2e/suite_test.go
alexander-demicev: it represents a management cluster and is created by Rancher

PR #134, opened 2023-04-18
🌱 Bump Go version and other dependencies

furkatgofurov7:
**What this PR does:**

- Uplifts Go version used in GHA workflows from 1.19 to 1.20
- Bumps go modules in root go.mod


go.mod
```diff
@@ -18,39 +18,41 @@ require (
 	github.com/rancher/wrangler-api v0.6.1-0.20200427172631-a7c2f09b783e
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.2
-	k8s.io/api v0.26.1
-	k8s.io/apiextensions-apiserver v0.26.1
-	k8s.io/apimachinery v0.26.1
-	k8s.io/apiserver v0.26.1
+	k8s.io/api v0.27.1
```
alexander-demicev, line 21: I think it's better to keep k8s dependencies aligned with core rancher https://github.com/rancher/rancher/blob/release/v2.7/go.mod#L25
Some packages from this operator are imported there so it might be better to share versions.

furkatgofurov7: Thanks for review, so we need to downgrade it?

alexander-demicev: I think so it might be easier for us to manage dependencies this way

furkatgofurov7: Done

richardcase (approved):

PR #138, opened 2023-04-20
Allow nil node role in cluster spec

alexander-demicev:
Allow nil node role in cluster spec will result in a node role being generated by the controller https://github.com/rancher/eks-operator/blob/master/controller/nodegroup.go#L261. Currently `""` is the only allowed option for generating a new node role but from our code logic point of view it shouldn't matter because we already treat empty strings and nil values the same.
I also made this field nil in our e2e so we can capture this case too.

https://github.com/rancher/rancher/issues/41123


furkatgofurov7 (commented):
/lgtm

a-blender (commented):
LGTM! This is a critical update for the next OOB charts release, thank you

@alexander-demicev 
* I also noticed this PR had only one reviewer requested. Please make sure all PRs have 2 approvals before merging, with at least one from an area SME
* I can only make comments not approve this for some reason. Anyone know why that is?

PR #142, opened 2023-04-24
Disable psp when installing rancher in e2e

alexander-demicev:
After the recent k8s version bump, we have to disable psp or rancher helm chart won't install.
```
"Error: INSTALLATION FAILED: execution error at (rancher/templates/validate-psp-install.yaml:4:5): The target cluster does not have the PodSecurityPolicy API resource. Please disable PSPs in this chart before proceeding.\n"
```


furkatgofurov7 (dismissed):
oops, forgot that, thanks!

furkatgofurov7 (approved):

PR #143, opened 2023-04-24
🌱 Enable more linters

furkatgofurov7:
This PR:

- Enables more linters in the .golangci.yaml to have a better aligned package import aliases
- Fixes all lint errors as a consequence of newly added linters


richardcase (approved):

alexander-demicev (approved):
thanks for fixing import aliases

Issue #106, opened 2023-02-14
Out of tree EBS CSI driver needed for EKS v1.23


richardcase:
With EKS v1.23 and higher it requires the use of the out of tree EBS CSI driver. If this isn't available then and PVC will not be bound and pods using the volumes will stay in pending.

This is a problem for creating new clusters and for upgrading clusters from 1.22

Docs: https://docs.amazonaws.cn/en_us/eks/latest/userguide/ebs-csi.html

**PR's**
- [x] https://github.com/rancher/eks-operator/pull/725
- [x] https://github.com/rancher/eks-operator/pull/724

sanjay920:
the aws docs are quite confusing.I added a wiki in the opni repo that our engineers use to get around this problem if it comes in handy https://github.com/rancher/opni/wiki/Install-AWS-EBS-CSI-driver

richardcase:
Thanks @sanjay920

richardcase:
A support utility has been created to handle this until we can include the functionality in eks-operator.

When including in the operator should we automatically enable the EBS CSI addon? Doing this will require that the IAM permissions that Rancher uses will need to be updated (including the docs). Or should it be an opt-in by annotating the namespace?

richardcase:
From discussion with @salasberryfin  & @mjura we thought this would be the best approach:

- Installation of the EBS CSI driver will be optional (as is the case when using the AWS console)
- We will add a new field to `EKSClusterConfigSpec` to indicate that the driver should be installed
- If the field is true we will do all the steps to enable the driver
- Update the EKS create cluster UI to have an option to set this new field. (optionally) In the UI we can also warn the user that additional permissions are required.
- Update the Rancher docs around the minimum EKS permissions: https://ranchermanager.docs.rancher.com/v2.5/reference-guides/amazon-eks-permissions/minimum-eks-permissions

salasberryfin:
On the `eks-operator` side, the issue can be divided in the following tasks:

- [ ] Add new boolean field to `EKSClusterConfigSpec`
- [ ] Implement logic to create OIDC provider and IAM role and install EKS addon
- [ ] Unit tests
- [ ] e2e tests

salasberryfin:
PR was merged and this is now waiting for rancher/dashboard#9043 to be implemented.

richardcase:
@salasberryfin - we'll also need to make an update to the Rancher Docs so that the new permissions are included.

salasberryfin:
[...]

PR #170, opened 2023-06-01
Out of tree EBS CSI Driver

salasberryfin:
Issue #106 

## Problem

With EKS v1.23 and higher it requires the use of the out of tree EBS CSI driver. If this isn't available then and PVC will not be bound and pods using the volumes will stay in pending.

This is a problem for creating new clusters and for upgrading clusters from 1.22

## Solution

The solution is based on the support utility rancherlabs/support-tools#217 which has been used as a workaround while the functionality is added to the operator.

- A boolean field is added to `EKSClusterConfigSpec` [EBSCSIDriver](https://github.com/salasberryfin/eks-operator/blob/46750b97cb4bb45019825cb2a90536bceca46854/pkg/apis/eks.cattle.io/v1/types.go#L46). This will require to update the EKS create cluster UI to have an option to set this new spec. 
- If set to true, the EBS CSI driver for EKS will be installed after the cluster is created.

### Installation procedure
- [Create OIDC provider for the cluster](https://github.com/salasberryfin/eks-operator/blob/46750b97cb4bb45019825cb2a90536bceca46854/pkg/eks/create.go#L530). This requires obtaining the thumbprint for the provider, which is a SHA1 of the certificate [see notes](#Notes).
- [Create IAM role](https://github.com/salasberryfin/eks-operator/blob/46750b97cb4bb45019825cb2a90536bceca46854/pkg/eks/create.go#L600) with the managed policy `AmazonEBSCSIDriverPolicy` which grants permissions to interact with EBS. The role allows `AssumeRoleWithWebIdentity` to the OIDC provider. (optionally) In the UI we can also warn the user that additional permissions are required. This role is created as a new [CloudFormation stack](https://github.com/salasberryfin/eks-operator/blob/46750b97cb4bb45019825cb2a90536bceca46854/templates/template.go#L260) `<cluster-name>-ebs-csi-driver-role`.
- [Install EBS add-on](https://github.com/salasberryfin/eks-operator/blob/46750b97cb4bb45019825cb2a90536bceca46854/pkg/eks/create.go#L634)

The cluster can be updated to install the add-on after it is created.

## Notes
- `golangci` prevents from using SHA1 for being a "weak cryptographic primitive", which required to add an [exclude rule].(https://github.com/salasberryfin/eks-operator/blob/46750b97cb4bb45019825cb2a90536bceca46854/.golangci.yml#L89) 
- The CloudFormation stack that creates the driver role is [configured to be removed](https://github.com/salasberryfin/eks-operator/blob/46750b97cb4bb45019825cb2a90536bceca46854/controller/eks-cluster-config-handler.go#L251) as other stacks associated with the cluster.
- Extra permissions need to be added to the IAM role if the user encrypts EBS volumes with a custom KMS key.
- Tests (and mocks) have been added for the new functionality.

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -447,22 +447,6 @@ func (h *Handler) create(config *eksv1.EKSClusterConfig, awsSVCs *awsServices) (
 		}
 	}
 
-	if aws.BoolValue(config.Spec.EBSCSIDriver) {
-		logrus.Infof("EBSCSIDriver is set to true: %v", config.Spec.EBSCSIDriver)
-		ebsCSIDriverInput := awsservices.EnableEBSCSIDriverInput{
-			EKSService: awsSVCs.eks,
-			IAMService: awsSVCs.iam,
-			CFService:  awsSVCs.cloudformation,
-			Config:     config,
-			// TODO: make this configurable?
-			AddonVersion: "latest",
-		}
-		err := awsservices.EnableEBSCSIDriver(ebsCSIDriverInput)
-		if err != nil {
-			return config, fmt.Errorf("error enabling EBS CSI driver: %w", err)
-		}
-	}
-
 	// If a user edits a cluster at the exact right (or wrong) time, then the
 	// `UpdateStatus` call may produce a conflict and will error. When the
 	// controller re-enters the create function, it will try to verify that a
@@ -1263,7 +1247,7 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
 				Config:       config,
 				AddonVersion: "latest",
 			}
-			if err := awsservices.EnableEBSCSIDriver(ebsCSIDriverInput); err != nil {
+			if err := awsservices.EnableEBSCSIDriver(&ebsCSIDriverInput); err != nil {
 				return config, fmt.Errorf("error enabling ebs csi driver addon: %w", err)
 			}
 		}
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -252,6 +252,14 @@ func (h *Handler) OnEksConfigRemoved(_ string, config *eksv1.EKSClusterConfig) (
 		}
 	}
 
+	if aws.BoolValue(config.Spec.EBSCSIDriver) == true {
+		logrus.Infof("deleting ebs csi driver role for config [%s]", config.Name)
+		err = deleteStack(h.awsServices.cloudformation, getEBSCSIDriverRoleStackName(config.Spec.DisplayName), getEBSCSIDriverRoleStackName(config.Spec.DisplayName))
+		if err != nil {
+			return config, fmt.Errorf("error ebs csi driver role stack: %v", err)
+		}
+	}
+
 	if aws.StringValue(config.Spec.ServiceRole) == "" {
 		logrus.Infof("deleting service role for config [%s]", config.Name)
 		err = deleteStack(awsSVCs.cloudformation, getServiceRoleName(config.Spec.DisplayName), getServiceRoleName(config.Spec.DisplayName))
@@ -1244,6 +1252,28 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
 		return h.enqueueUpdate(config)
 	}
 
+	// check if ebs csi driver needs to be enabled
+	if aws.BoolValue(config.Spec.EBSCSIDriver) {
+		installedArn, err := awsservices.CheckEBSAddon(h.awsServices.eks, config)
+		if err != nil {
+			return nil, fmt.Errorf("error checking if ebs csi driver addon is installed: %v", err)
+		}
+		if installedArn == "" {
+			logrus.Infof("enabling [ebs csi driver add-on] for cluster [%s]", config.Spec.DisplayName)
+			ebsCSIDriverInput := awsservices.EnableEBSCSIDriverInput{
+				EKSService:   h.awsServices.eks,
+				IAMService:   h.awsServices.iam,
+				CFService:    h.awsServices.cloudformation,
+				Config:       config,
+				AddonVersion: "latest",
+			}
+			err := awsservices.EnableEBSCSIDriver(ebsCSIDriverInput)
+			if err != nil {
+				return config, fmt.Errorf("error enabling ebs csi driver addon: %w", err)
+			}
+		}
+	}
+
 	// no new updates, set to active
 	if config.Status.Phase != eksConfigActivePhase {
 		logrus.Infof("cluster [%s] finished updating", config.Name)
@@ -1333,6 +1363,10 @@ func (h *Handler) enqueueUpdate(config *eksv1.EKSClusterConfig) (*eksv1.EKSClust
 
 func getVPCStackName(name string) string {
 	return name + "-eks-vpc"
+}
+
+func getEBSCSIDriverRoleStackName(name string) string {
+	return name + "-ebs-csi-driver-role"
 }
 
 func getServiceRoleName(name string) string {
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -252,33 +252,29 @@ func (h *Handler) OnEksConfigRemoved(_ string, config *eksv1.EKSClusterConfig) (
 		}
 	}
 
-	if aws.BoolValue(config.Spec.EBSCSIDriver) == true {
+	if aws.BoolValue(config.Spec.EBSCSIDriver) {
 		logrus.Infof("deleting ebs csi driver role for config [%s]", config.Name)
-		err = deleteStack(h.awsServices.cloudformation, getEBSCSIDriverRoleStackName(config.Spec.DisplayName), getEBSCSIDriverRoleStackName(config.Spec.DisplayName))
-		if err != nil {
+		if err := deleteStack(awsSVCs.cloudformation, getEBSCSIDriverRoleStackName(config.Spec.DisplayName), getEBSCSIDriverRoleStackName(config.Spec.DisplayName)); err != nil {
 			return config, fmt.Errorf("error ebs csi driver role stack: %v", err)
 		}
 	}
 
 	if aws.StringValue(config.Spec.ServiceRole) == "" {
 		logrus.Infof("deleting service role for config [%s]", config.Name)
-		err = deleteStack(awsSVCs.cloudformation, getServiceRoleName(config.Spec.DisplayName), getServiceRoleName(config.Spec.DisplayName))
-		if err != nil {
+		if err := deleteStack(awsSVCs.cloudformation, getServiceRoleName(config.Spec.DisplayName), getServiceRoleName(config.Spec.DisplayName)); err != nil {
 			return config, fmt.Errorf("error deleting service role stack: %v", err)
 		}
 	}
 
 	if len(config.Spec.Subnets) == 0 {
 		logrus.Infof("deleting vpc, subnets, and security groups for config [%s]", config.Name)
-		err = deleteStack(awsSVCs.cloudformation, getVPCStackName(config.Spec.DisplayName), getVPCStackName(config.Spec.DisplayName))
-		if err != nil {
+		if err := deleteStack(awsSVCs.cloudformation, getVPCStackName(config.Spec.DisplayName), getVPCStackName(config.Spec.DisplayName)); err != nil {
 			return config, fmt.Errorf("error deleting vpc stack: %v", err)
 		}
 	}
 
 	logrus.Infof("deleting node instance role for config [%s]", config.Name)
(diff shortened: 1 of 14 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -248,6 +248,14 @@ func (h *Handler) OnEksConfigRemoved(_ string, config *eksv1.EKSClusterConfig) (
 		}
 	}
 
+	if aws.BoolValue(config.Spec.EBSCSIDriver) == true {
```
alexander-demicev, line 251:
```suggestion
	if aws.BoolValue(config.Spec.EBSCSIDriver) {
```

controller/eks-cluster-config-handler.go
```diff
@@ -248,6 +248,14 @@ func (h *Handler) OnEksConfigRemoved(_ string, config *eksv1.EKSClusterConfig) (
...
+		logrus.Infof("deleting ebs csi driver role for config [%s]", config.Name)
+		err = deleteStack(h.awsServices.cloudformation, getEBSCSIDriverRoleStackName(config.Spec.DisplayName), getEBSCSIDriverRoleStackName(config.Spec.DisplayName))
```
alexander-demicev, line 253:
```suggestion
		if err := deleteStack(h.awsServices.cloudformation, getEBSCSIDriverRoleStackName(config.Spec.DisplayName), getEBSCSIDriverRoleStackName(config.Spec.DisplayName)); err != {
```
I know inline ifs are not consistently used in the operator code but it's a golang idiomatic way and we can start doing it

controller/eks-cluster-config-handler.go
```diff
@@ -1200,6 +1208,28 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
 		return h.enqueueUpdate(config)
 	}
 
+	// check if ebs csi driver needs to be enabled
+	if aws.BoolValue(config.Spec.EBSCSIDriver) {
+		installedArn, err := awsservices.CheckEBSAddon(h.awsServices.eks, config)
+		if err != nil {
+			return nil, fmt.Errorf("error checking if ebs csi driver addon is installed: %v", err)
```
alexander-demicev, line 1215:
```suggestion
			return nil, fmt.Errorf("error checking if ebs csi driver addon is installed: %w", err)
```

pkg/eks/create.go
```diff
@@ -463,3 +474,177 @@ func getParameterValueFromOutput(key string, outputs []*cloudformation.Output) s
...
+func EnableEBSCSIDriver(opts EnableEBSCSIDriverInput) error {
+	oidcID, err := configureOIDCProvider(opts.IAMService, opts.EKSService, opts.Config)
+	if err != nil {
+		return fmt.Errorf("could not configure oidc provider: %v", err)
+	}
+	roleArn, err := createEBSCSIDriverRole(opts.CFService, opts.Config, oidcID)
+	if err != nil {
+		return fmt.Errorf("could not create ebs csi driver role: %v", err)
+	}
+	_, err = installEBSAddon(opts.EKSService, opts.Config, roleArn, opts.AddonVersion)
```
alexander-demicev, line 498:
```suggestion
	if _, err := installEBSAddon(opts.EKSService, opts.Config, roleArn, opts.AddonVersion); err != nil {
```

pkg/eks/create.go
```diff
@@ -463,3 +474,177 @@ func getParameterValueFromOutput(key string, outputs []*cloudformation.Output) s
...
+// the ARN of the add-on. If it is not, it will return an empty string. Otherwise, it will return an error
+func CheckEBSAddon(eksService services.EKSServiceInterface, config *eksv1.EKSClusterConfig) (string, error) {
+	input := eks.DescribeAddonInput{
+		AddonName:   aws.String(ebsCSIAddonName),
+		ClusterName: aws.String(config.Spec.DisplayName),
+	}
+
+	output, err := eksService.DescribeAddon(&input)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
```
alexander-demicev, line 516: you can use `errors.As` here https://pkg.go.dev/errors, it makes the code a bit more readable

pkg/eks/create.go
```diff
@@ -463,3 +474,177 @@ func getParameterValueFromOutput(key string, outputs []*cloudformation.Output) s
...
+	if err != nil {
+		return "", err
+	}
+	clusterOutput, err := eksService.DescribeCluster(&eks.DescribeClusterInput{
+		Name: aws.String(config.Spec.DisplayName),
+	})
+	if err != nil {
+		return "", err
+	}
+	id := path.Base(*clusterOutput.Cluster.Identity.Oidc.Issuer)
```
alexander-demicev, line 541: I know it's very unlikely to happen but can we check that `clusterOutput` is not nil?

pkg/eks/create.go
```diff
@@ -463,3 +474,177 @@ func getParameterValueFromOutput(key string, outputs []*cloudformation.Output) s
...
+	if version != "latest" {
+		input.AddonVersion = aws.String(version)
+	}
+
+	addonOutput, err := eksService.CreateAddon(&input)
+	if err != nil {
+		return "", err
+	}
+
+	return *addonOutput.Addon.AddonArn, nil
```
alexander-demicev, line 649: same as above, should we add nil check?

pkg/eks/create.go
```diff
@@ -463,3 +474,177 @@ func getParameterValueFromOutput(key string, outputs []*cloudformation.Output) s
...
+	if err != nil {
+		return fmt.Errorf("failed to install ebs csi driver addon: %v", err)
+	}
+
+	return nil
+}
+
+// CheckEBSAddon checks if the EBS CSI driver add-on is installed. If it is, it will return
+// the ARN of the add-on. If it is not, it will return an empty string. Otherwise, it will return an error
+func CheckEBSAddon(eksService services.EKSServiceInterface, config *eksv1.EKSClusterConfig) (string, error) {
```
alexander-demicev, line 508: looks like this function belongs to `get.go`

alexander-demicev (commented):
The PR looks great! Thanks a lot, I left some minor comments.

controller/eks-cluster-config-handler.go
```diff
@@ -1200,6 +1208,28 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
...
+		if installedArn == "" {
+			logrus.Infof("enabling [ebs csi driver add-on] for cluster [%s]", config.Spec.DisplayName)
+			ebsCSIDriverInput := awsservices.EnableEBSCSIDriverInput{
+				EKSService:   h.awsServices.eks,
+				IAMService:   h.awsServices.iam,
+				CFService:    h.awsServices.cloudformation,
+				Config:       config,
+				AddonVersion: "latest",
+			}
+			err := awsservices.EnableEBSCSIDriver(ebsCSIDriverInput)
```
mjura, line 1226:
```suggestion
			if err := awsservices.EnableEBSCSIDriver(ebsCSIDriverInput); err != nil {
```

mjura (changes requested):

pkg/eks/create.go
salasberryfin: I was thinking about how could I implement this with `errors.As` but I don't know how to get rid of the `err.(awserr.Error)` to cast it to a `awserr` that I can then compare against `eks.ErrCodeResourceNotFoundException`, which is just an error code represented as string.

```go
var genericAWSErr awserr.Error
if errors.As(err, &genericAWSErr) {
	if aerr, _ := err.(awserr.Error); aerr.Code() == eks.ErrCodeResourceNotFoundException {
		// resource not found
	}
}
```

Is there a way I could make it easier and more readable?

salasberryfin:
Thanks everyone for the reviews. Just addressed the proposed changes in the last commit.

salasberryfin:
Need to update AWS clients when #176 is merged.

pkg/apis/eks.cattle.io/v1/types.go
```diff
@@ -43,6 +43,7 @@ type EKSClusterConfigSpec struct {
 	KmsKey                 *string           `json:"kmsKey" norman:"noupdate,pointer"`
 	PublicAccess           *bool             `json:"publicAccess"`
 	PrivateAccess          *bool             `json:"privateAccess"`
+	EBSCSIDriver           *bool             `json:"ebsCSIDriver"`
```
richardcase, line 46: As we are adding it on the CRD spec and making optional (i.e. not auto installing for 1.23 or higher clusters) we'll need a UI change to expose this option.

pkg/eks/create.go
```diff
@@ -463,3 +474,158 @@ func getParameterValueFromOutput(key string, outputs []*cloudformation.Output) s
...
+	if err != nil {
+		return "", err
+	}
+	buf := &bytes.Buffer{}
+	if execErr := tmpl.Execute(buf, templateData); execErr != nil {
+		return "", err
+	}
+	finalTemplate := buf.String()
+
+	output, err := CreateStack(CreateStackOptions{
```
richardcase, line 596: Would be great to not introduce any more CloudFormation....but understand why as all the other roles are created using CF

richardcase (commented):
Looking good to me.

pkg/eks/create.go
alexander-demicev:
```golang
var genericAWSErr awserr.Error
if errors.As(err, &genericAWSErr) && aerr.Code() == eks.ErrCodeResourceNotFoundException {
}
```
can something like this work?

pkg/apis/eks.cattle.io/v1/types.go
salasberryfin: I've created rancher/dashboard#9043 to request this UI change.

pkg/eks/create.go
salasberryfin: I'm happy to move this to the SDK if that's better. I just used CloudFormation to align with other AWS resources created in the repo and so I could use the same removal process as well.

controller/eks-cluster-config-handler.go
```diff
@@ -443,6 +447,22 @@ func (h *Handler) create(config *eksv1.EKSClusterConfig, awsSVCs *awsServices) (
...
+		logrus.Infof("EBSCSIDriver is set to true: %v", config.Spec.EBSCSIDriver)
+		ebsCSIDriverInput := awsservices.EnableEBSCSIDriverInput{
+			EKSService: awsSVCs.eks,
+			IAMService: awsSVCs.iam,
+			CFService:  awsSVCs.cloudformation,
+			Config:     config,
+			// TODO: make this configurable?
+			AddonVersion: "latest",
+		}
+		err := awsservices.EnableEBSCSIDriver(ebsCSIDriverInput)
```
alexander-demicev, line 460:
```suggestion
		if err := awsservices.EnableEBSCSIDriver(ebsCSIDriverInput); if err != nil {
```

pkg/eks/create.go
```diff
@@ -463,3 +474,158 @@ func getParameterValueFromOutput(key string, outputs []*cloudformation.Output) s
...
+	EKSService   services.EKSServiceInterface
+	IAMService   services.IAMServiceInterface
+	CFService    services.CloudFormationServiceInterface
+	Config       *eksv1.EKSClusterConfig
+	AddonVersion string
+}
+
+// EnableEBSCSIDriver manages the installation of the EBS CSI driver for EKS, including the
+// creation of the OIDC Provider, the IAM role and the validation and installation of the EKS add-on
+func EnableEBSCSIDriver(opts EnableEBSCSIDriverInput) error {
```
alexander-demicev, line 489:
```suggestion
func EnableEBSCSIDriver(opts *EnableEBSCSIDriverInput) error {
```
Please, make it a pointer, we changed other functions as well. Sometimes you need to modify, for example, `EKSClusterConfig` that is passed here.

alexander-demicev (commented):
just couple of minor comments, everything else looks great! Thanks a lot for working on it.

alexander-demicev (approved):

alexander-demicev (approved):

richardcase (commented):
@salasberryfin - looks good to me. One question I have is, as we are installing the **latest** version of the addon, let's say that is version 2. What happens when a new version is released (i.e. version 3)?

salasberryfin:
@richardcase that's a good point. As of now, the version cannot be updated from Rancher with this implementation. Depending on what we want to allow the user to do, we could let them choose what version of the addon they would like to be installed. This would require the UI to show a list of versions (similar to what's shown for subnets). Then, after the cluster has been provisioned, this could be changed in the configuration.

salasberryfin:
> @salasberryfin - looks good to me. One question I have is, as we are installing the **latest** version of the addon, let's say that is version 2. What happens when a new version is released (i.e. version 3)?

@richardcase would it make sense to merge this as an initial implementation and then add support to updates and specific add-on versions as an improvement to this feature? We can add a follow up issue to the backlog.

richardcase:
> @richardcase would it make sense to merge this as an initial implementation and then add support to updates and specific add-on versions as an improvement to this feature? We can add a follow up issue to the backlog.

@salasberryfin - that seems reasonable to me. Especially, as the addon can be updated via the AWS console.

mjura (approved):
LGTM

PR #149, opened 2023-05-02
Bring back functions that are used in rancher

alexander-demicev:
Some functions are used in rancher/rancher so they have to be public. We added API diff workflow to make sure that it won't happen again.

PR #176, opened 2023-06-05
Fix AWS clients initialization

alexander-demicev:
This PR fixes couple of issues:
- Remove AWS clients from the handler structure, currently, we reassign clients on each reconciliation. It hasn't been an issue but is not an ideal solution.
- Initialize AWS clients also in `OnEksConfigRemoved`
- Make arguments for AWS service calls pointers because they sometimes have to modify input objects.

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -48,7 +48,6 @@ type Handler struct {
 	eksEnqueue      func(namespace, name string)
 	secrets         wranglerv1.SecretClient
 	secretsCache    wranglerv1.SecretCache
-	awsServices     awsServices
 }
 
 type awsServices struct {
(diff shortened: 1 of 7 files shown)


richardcase (commented):
Generally LGTM.

Do we need to worry about the apidiff warnings? Does Rancher depend on any of these exported functions?

alexander-demicev:
No, these are only used in eks operator

salasberryfin (approved):
LGTM. Once this is merged, #170 will also need to be updated to use AWS clients properly.

mjura (approved):
LGTM

furkatgofurov7:
> Does Rancher depend on any of these exported functions?

I had the same understanding, and in the past we had issues with that and https://github.com/rancher/eks-operator/pull/149 was introduced

alexander-demicev:
We can also move them to `internal` package to ensure that no one imports them as these functions are not intended to be used outside of the operator.

richardcase:
> No, these are only used in eks operator

Like @furkatgofurov7 i had that other issue we encountered in mind. Also agree @alexander-demicev , in a separate PR we can think about moving stuff to `internal`.

richardcase (approved):

furkatgofurov7 (approved):

PR #177, opened 2023-06-05
Fix chart update workflow

alexander-demicev:
If we're adding a new version to [release.yaml](https://github.com/rancher/charts/blob/dev-v2.7/release.yaml) there 2 things the script can do:
- If eks-operator record is present in the list, it should add new version to the list, currently `sed` instruction for it is broken because it doesn't add proper indentation.
- If eks-operator record is not present, the script creates it. In the current implementation, `if grep -qv "rancher-eks-operator:" release.yaml; then` doesn't work as expected so I replaced it with a new command.

Result of the test run: https://github.com/rancher/charts/pull/2706


richardcase (approved):

furkatgofurov7 (approved):

furkatgofurov7:
Should we cherry pick the same for aks-operator as well?

mjura (approved):
LGTM

Issue #85, opened 2023-02-03
(SURE-5259) EKS Provision Failure w/ Rancher2 Terraform Provider


kkaempf:
Seems very similar to SURE-4066

#### Issue description:
Can't provision an EKS downstream cluster using the Rancher2 TF provider. This is the error we see in the Rancher UI on the cluster as well as inside the eks-config-operator pod logs (it is spammed continuously):
```
time="2022-09-13T23:43:46Z" level=error msg="error syncing 'cattle-global-data/c-ksl6p': handler eks-controller: InvalidParameterException: Launch template details can't be null for Custom ami type node group\n{\n  RespMetadata: {\n    StatusCode: 400,\n    RequestID: \"eca6e2f1-42d5-411f-b2fc-716404d09d13\"\n  },\n  Message_: \"Launch template details can't be null for Custom ami type node group\"\n}, requeuing" 
```
Cloud Trail on the AWS backned is reporting this error for the EventType: UpdateNodegroupVersion

#### Error code
InvalidParameterException

Event Record.
```
{
    "eventVersion": "1.08",
    "userIdentity": {
        "type": "IAMUser",
        "principalId": "AIDA4SBL6SADYCHEP64QO",
        "arn": "arn:aws:iam::863380606983:user/srvamr-btcsapid",
        "accountId": "1111111111111",
        "accessKeyId": "XXXXXXXXXXXXXXXXXXXXXXXX",
        "userName": "srvamr-btcsapid"
    },
    "eventTime": "2022-09-02T19:42:50Z",
    "eventSource": "eks.amazonaws.com",
    "eventName": "UpdateNodegroupVersion",
    "awsRegion": "us-east-1",
    "sourceIPAddress": "148.168.40.5",
    "userAgent": "aws-sdk-go/1.36.7 (go1.16.4; linux; amd64)",
    "errorCode": "InvalidParameterException",
    "requestParameters": {
        "nodegroupName": "pdcs-dev1d-harim-v2-090122-ng1",
        "clientRequestToken": "D9CB6CAB-3459-4E09-89F0-DE4CF3BB6CAE",
        "name": "pdcs-dev1d-harim-v2-090122",
        "version": "1.21"
    },
    "responseElements": {
        "message": "Launch template details can't be null for Custom ami type node group"
    },
    "requestID": "9b15e35e-8349-4f9f-9586-62e94e253308",
    "eventID": "6bbdb28c-3b56-4011-90a6-27f791fdf035",
[...]

PR #186, opened 2023-06-15
Fix error updating node groups created with a custom launch template

salasberryfin:
## Problem

The issue this PR fixes comes from the way version updates work in EKS node groups:
- "Standard" node groups, which are those created without a custom launch template, can be updated to a newer version by simply using the AWS SDK `UpdateNodegroupVersion` with the specific version the node group is expected to update to.
- Node groups created from a custom launch template (which use a custom AMI) cannot just be updated to a newer version. The only way to update these node groups is to create a new version of the launch template and point this node group to the newer version.

The eks-operator does not manage these different behaviors depending on how node groups are originally created, and hence submit a malformed request to `UpdateNodegroupVersion` in the specific case of a custom launch template node group, causing the error reported in #85 .

## Solution

Version updates for node groups should follow:
- Node groups without a custom launch template can be updated by just specifying the new version, which aligns with the option shown in Rancher UI.
- Node groups created from a custom launch template cannot be updated to a new version. The correct way to update these node groups is to update the launch template version (which is possible from Rancher UI).

With this change, we keep track of whether the node group is created with a Rancher-managed launch template or a custom launch template (`rancherManagedLaunchTemplate`). Now, in addition to a new version, we also need this value to be set to true before submitting a request to `UpdateNodegroupVersion`. Modifying the launch template version would be the only way to trigger an update on `rancherManagedLaunchTemplate = false` node groups.

## Notes

Rancher UI shows the option of updating the node group version when the cluster uses a newer version.

Considering we have an **EKS cluster using v1.24** and two node groups:
- `ng1` is a basic node group with a Rancher-managed launch template: **v1.23**
- `ng2` is a node group created from a custom launch template: **v1.23**

As a user editing this cluster, Rancher allows to tick a box to update each node group to v1.24. This is valid for `ng1`, which will be successfully updated, but is not correct for `ng2`, since we cannot update its version unless there's a new launch template version that uses v1.24. From the UI we can use the proper method for `ng2`, which is specifying a different launch template version. Ideally, we should remove the option of version update for node groups that were originally created with a custom launch template.

The following snapshot shows the correct (green) and incorrect (red) way to update `ng2`.

image

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -1140,12 +1140,15 @@ func (h *Handler) updateUpstreamClusterState(upstreamSpec *eksv1.EKSClusterConfi
 			ClusterName:   aws.String(config.Spec.DisplayName),
 		}
 
+		// rancherManagedLaunchTemplate is true if user did not specify a custom launch template
+		rancherManagedLaunchTemplate := false
 		if upstreamNg.LaunchTemplate != nil {
 			upstreamTemplateVersion := aws.Int64Value(upstreamNg.LaunchTemplate.Version)
 			var err error
 			lt := ng.LaunchTemplate
 
 			if lt == nil && config.Status.ManagedLaunchTemplateID == aws.StringValue(upstreamNg.LaunchTemplate.ID) {
+				rancherManagedLaunchTemplate = true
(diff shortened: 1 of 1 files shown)


mjura (approved):
LGTM

furkatgofurov7 (approved):

Issue #203, opened 2023-07-11
Document EKSClusterConfig API type


richardcase:
## SURE-5202

We need to document the [EKSClusterConfig](https://github.com/rancher/eks-operator/blob/master/pkg/apis/eks.cattle.io/v1/types.go#L26C6-L26C22) API type including all child file and structs with:

- documentation comments
- kubebuilder validation annotations

### PR:

- [x] https://github.com/rancher/eks-operator/pull/225

PR #225, opened 2023-08-26
Document EKSClusterConfig API

mjura:
Issue: https://github.com/rancher/eks-operator/issues/203

### DIFF
--- a/pkg/apis/eks.cattle.io/v1/types.go
+++ b/pkg/apis/eks.cattle.io/v1/types.go
@@ -22,6 +22,10 @@ import (
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:name="DisplayName",type="string",JSONPath=".spec.clusterName"
+// +kubebuilder:printcolumn:name="KubernetesVersion",type="string",JSONPath=".spec.kubernetesVersion"
+// +kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase"
+// +kubebuilder:printcolumn:name="FailureMessage",type="string",JSONPath=".status.failureMessage"
 
 type EKSClusterConfig struct {
 	metav1.TypeMeta   `json:",inline"`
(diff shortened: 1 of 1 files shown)


pkg/apis/eks.cattle.io/v1/types.go
```diff
@@ -33,63 +33,111 @@ type EKSClusterConfig struct {
...
+	// AmazonCredentialSecret is the name of the secret containing the Amazon credentials.
+	AmazonCredentialSecret string `json:"amazonCredentialSecret"`
+	// DisplayName is the name of the cluster to be displayed in the UI.
+	DisplayName string `json:"displayName" norman:"noupdate"`
+	// Region is the AWS region to create the cluster in.
+	Region string `json:"region" norman:"noupdate"`
+	// Imported is true if the cluster was imported.
+	Imported bool `json:"imported" norman:"noupdate"`
+	// KubernetesVersion is the version of Kubernetes to use.
+	KubernetesVersion *string `json:"kubernetesVersion" norman:"pointer"`
```
richardcase, line 45: Normally with kubebuilder/controller runtime a pointer indicates that a field is optional. If thats the case for the pointer values here we can `+optional`

pkg/apis/eks.cattle.io/v1/types.go
```diff
@@ -33,63 +33,111 @@ type EKSClusterConfig struct {
...
+	NodeGroups []NodeGroup `json:"nodeGroups"`
 }
 
 type EKSClusterConfigStatus struct {
-	Phase                         string            `json:"phase"`
-	VirtualNetwork                string            `json:"virtualNetwork"`
-	Subnets                       []string          `json:"subnets"`
-	SecurityGroups                []string          `json:"securityGroups"`
-	ManagedLaunchTemplateID       string            `json:"managedLaunchTemplateID"`
+	// Phase is the current lifecycle phase of the cluster.
```
richardcase, line 73: It would also be good to add doc strings to the struct definitions.

richardcase (commented):
Thanks @mjura 👍  Just a few minor comments.

pkg/apis/eks.cattle.io/v1/types.go
mjura: fixed

pkg/apis/eks.cattle.io/v1/types.go
```diff
@@ -33,63 +37,156 @@ type EKSClusterConfig struct {
...
+	// +kubebuilder:validation:UniqueItems:=true
+	Tags map[string]string `json:"tags"`
+	// SecretsEncryption is true if secrets should be encrypted.
+	// +optional
+	SecretsEncryption *bool `json:"secretsEncryption" norman:"noupdate"`
+	// KmsKey is the KMS key to use for encryption.
+	// +optional
+	KmsKey *string `json:"kmsKey" norman:"noupdate,pointer"`
+	// PublicAccess is true if the cluster should be publicly accessible.
+	// +kubebuilder:validation:Required
```
richardcase, line 66: Are these required, even though they are pointer values?

richardcase (commented):
@mjura - are you sure that everything marked as "required" is actually required? The reason i ask is that a lof of the types are pointers which can denote something being optional.

pkg/apis/eks.cattle.io/v1/types.go
mjura: fixed

mjura:
> @mjura - are you sure that everything marked as "required" is actually required? The reason i ask is that a lof of the types are pointers which can denote something being optional.

I think that pointers should be verified, but let's stay consistent with Kubebuilder markers

salasberryfin (approved):

PR #232, opened 2023-08-29
fix: rancher panics when node group fails

salasberryfin:
## Description

This panic is caused by a node group created from the AWS Console using a malformed launch template. This was specifically detected while trying to use the default version of the Rancher-managed launch template, which only contains a placeholder string in the user data.


furkatgofurov7 (commented):
Do we have a unit test for the EKS cluster config handler to mock this behavior? Otherwise looks good to me.

mjura (approved):
lgtm

salasberryfin:
> Do we have a unit test for the EKS cluster config handler to mock this behavior? Otherwise looks good to me.

@furkatgofurov7 I see we don't have unit tests for `eks-cluster-config-handler.go` just yet, but I'll add an issue to track this.

PR #236, opened 2023-08-31
[v2.7] fix: rancher panics when node group fails

salasberryfin:
## Description

This panic is caused by a node group created from the AWS Console using a malformed launch template. This was specifically detected while trying to use the default version of the Rancher-managed launch template, which only contains a placeholder string in the user data.


mjura (approved):
lgtm

PR #247, opened 2023-09-04
[2.7] Sync go.mod with release-v2.8

mjura:


Danil-Grigorev (approved):

furkatgofurov7 (changes requested):
@mjura can you please add a PR description with information, i.e what issue this PR is solving / why we need it? Otherwise, sometimes it is hard to guess what are we trying with these kinds of PRs

furkatgofurov7:
Q: why we should sync v2.7 go mod with v2.8, as I think they are different versions of operators tracking corresponding versions of Rancher and they should have a diff between them, no?

mjura:
> Q: why we should sync v2.7 go mod with v2.8, as I think they are different versions of operators tracking corresponding versions of Rancher and they should have a diff between them, no?

Yes, I agree. It is backport from v2.8 to v2.7. Added comment to commit message

furkatgofurov7 (approved):
Not sure how these deps got bumped (we have [ignore](https://github.com/rancher/eks-operator/blob/release-v2.7/.github/dependabot.yaml#L19) for it in dependabot) to these versions though the changes in this PR makes sense to me. 
@mjura probably we could also decide to ignore `aws-sdk-go` bumps in the dependabot configuration, if it break the stuff and we want to do it manually.

salasberryfin (approved):

furkatgofurov7:
> rancher/rancher#41716 (comment)

or have a CI to test it when the bump PR is opened

PR #542, opened 2024-05-21
chore: make dependabot ignore all k8s updates

salasberryfin:

Dependabot is ignoring updates for k8s dependencies but only for major and minor versions. As k8s is updated manually in sync with Rancher core, we also need to ignore patch releases.

Issue #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed


mjura (approved):
LGTM

PR #548, opened 2024-05-23
fix: Replace unapproved GH Actions with approved ones

yiannistri:

Replace unapproved GH Actions with approved ones. To [enhance security](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), the 3rd-party GH Action used to install kind is pinned to a commit SHA. Note that dependabot will [update](https://github.com/dependabot/dependabot-core/issues/4691) both the commit SHA and the human-readable value next to it.

Issue https://github.com/rancher/highlander/issues/89

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [x] backport needed

### DIFF
--- a/.github/workflows/apidiff.yaml
+++ b/.github/workflows/apidiff.yaml
@@ -11,4 +11,5 @@ jobs:
     - uses: actions/setup-go@v5
       with:
         go-version: 1.21.x
-    - uses: joelanford/go-apidiff@main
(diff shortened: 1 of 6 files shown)


salasberryfin (approved):
Thanks @yiannistri

mjura (approved):

PR #597, opened 2024-06-11
Enhance logs

valaparthvi:
/kind enhancement

1. Use [cluster-name (id: cluster-id)] instead of either [cluster-name] or [cluster-id]
2. Log more information

Issue #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -14,18 +14,19 @@ import (
 	ekstypes "github.com/aws/aws-sdk-go-v2/service/eks/types"
 	"github.com/aws/aws-sdk-go-v2/service/iam"
 	"github.com/blang/semver"
-	eksv1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
-	awsservices "github.com/rancher/eks-operator/pkg/eks"
-	"github.com/rancher/eks-operator/pkg/eks/services"
-	ekscontrollers "github.com/rancher/eks-operator/pkg/generated/controllers/eks.cattle.io/v1"
-	"github.com/rancher/eks-operator/templates"
-	"github.com/rancher/eks-operator/utils"
 	wranglerv1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
+
+	eksv1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
+	awsservices "github.com/rancher/eks-operator/pkg/eks"
+	"github.com/rancher/eks-operator/pkg/eks/services"
+	ekscontrollers "github.com/rancher/eks-operator/pkg/generated/controllers/eks.cattle.io/v1"
+	"github.com/rancher/eks-operator/templates"
+	"github.com/rancher/eks-operator/utils"
 )
 
 const (
@@ -138,7 +139,7 @@ func (h *Handler) recordError(onChange func(key string, config *eksv1.EKSCluster
 		var recordErr error
 		config, recordErr = h.eksCC.UpdateStatus(config)
 		if recordErr != nil {
-			logrus.Errorf("Error recording ekscc [%s] failure message: %s", config.Name, recordErr.Error())
+			logrus.Errorf("Error recording ekscc [%s (id: %s)] failure message: %s", config.Spec.DisplayName, config.Name, recordErr.Error())
 		}
 		return config, err
 	}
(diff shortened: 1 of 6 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -335,7 +336,7 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
 		if _, ok := nodeGroupNames[aws.ToString(ng.NodegroupName)]; !ok {
 			nodeGroupNames[aws.ToString(ng.NodegroupName)] = struct{}{}
 		} else {
-			errs = append(errs, fmt.Sprintf("node group names must be unique within the [%s] cluster to avoid duplication", config.Name))
+			errs = append(errs, fmt.Sprintf("node group name %q must be unique within the [%s (id: %s)] cluster to avoid duplication", aws.ToString(ng.NodegroupName), config.Spec.DisplayName, config.Name))
```
mjura, line 339:
```suggestion
			errs = append(errs, fmt.Sprintf("node group names must be unique within the [%s (id: %s)] cluster to avoid duplication", config.Spec.DisplayName, config.Name))
```

controller/eks-cluster-config-handler.go
```diff
@@ -444,114 +445,114 @@ func (h *Handler) validateCreate(ctx context.Context, config *eksv1.EKSClusterCo
...
+					return fmt.Errorf(cannotBeNilError, "instanceType", *ng.NodegroupName, config.Spec.DisplayName, config.Name)
 				}
 			}
 			if ng.NodegroupName == nil {
-				return fmt.Errorf(cannotBeNilError, "name", *ng.NodegroupName, config.Name)
+				return fmt.Errorf(cannotBeNilError, "name", *ng.NodegroupName, config.Spec.DisplayName, config.Name)
 			}
 			if nodeP[*ng.NodegroupName] {
-				return fmt.Errorf("node group names must be unique within the [%s] cluster to avoid duplication", config.Name)
+				return fmt.Errorf("node group name [%s] must be unique within the [%s (id: %s)] cluster to avoid duplication", *ng.NodegroupName, config.Spec.DisplayName, config.Name)
```
mjura, line 511:
```suggestion
				return fmt.Errorf("node group names must be unique within the [%s (id: %s)] cluster to avoid duplication", config.Spec.DisplayName, config.Name)
```

controller/eks-cluster-config-handler.go
```diff
@@ -444,114 +445,114 @@ func (h *Handler) validateCreate(ctx context.Context, config *eksv1.EKSClusterCo
...
-				return fmt.Errorf(cannotBeNilError, "labels", *ng.NodegroupName, config.Name)
+				return fmt.Errorf(cannotBeNilError, "labels", *ng.NodegroupName, config.Spec.DisplayName, config.Name)
 			}
 			if ng.RequestSpotInstances == nil {
-				return fmt.Errorf(cannotBeNilError, "requestSpotInstances", *ng.NodegroupName, config.Name)
+				return fmt.Errorf(cannotBeNilError, "requestSpotInstances", *ng.NodegroupName, config.Spec.DisplayName, config.Name)
 			}
 			if ng.NodeRole == nil {
-				logrus.Warnf("nodeRole is not specified for nodegroup [%s] in cluster [%s], the controller will generate it", *ng.NodegroupName, config.Name)
+				logrus.Warnf("nodeRole is not specified for nodegroup [%s] in cluster [%s (id: %s)], the controller will generate it", *ng.NodegroupName, config.Spec.DisplayName, config.Name)
```
mjura, line 542:
```suggestion
				logrus.Warnf("nodeRole is not specified for nodegroup [%s] in cluster [%s (id: %s)], the controller will generate it", aws.ToString(ng.NodegroupName), config.Spec.DisplayName, config.Name)
```

pkg/eks/update.go
```diff
@@ -51,6 +53,9 @@ type UpdateResourceTagsOpts struct {
 func UpdateResourceTags(ctx context.Context, opts *UpdateResourceTagsOpts) (bool, error) {
 	updated := false
 	if updateTags := utils.GetKeyValuesToUpdate(opts.Tags, opts.UpstreamTags); updateTags != nil {
+		logrus.Infof("updating resource tags to %s for cluster [%s]", opts.Tags, opts.ClusterName)
+		logrus.Debugf("config: %s, upstream: %s", opts.Tags, opts.UpstreamTags)
```
mjura, line 57: I have doubts if this will work with `%s`

mjura (changes requested):
there is still place for improvements

pkg/eks/update.go
valaparthvi: %s works here since `opts.Tags` is a map[string]string, but it is better to change this to %v in any case, thank you for catching this!

controller/eks-cluster-config-handler.go
valaparthvi: To be consistent with gke-operator, and to have a more clear error, I have modified the error message to be `node group name [%s] is not unique within the ....`.

valaparthvi:
@mjura Can you take a look at this? I was hoping to merge it before the next release, so we can test it enough and catch any corner errors if any.

mjura (approved):

PR #724, opened 2024-08-01
Set information about CSI EBS driver

mjura:
Issue: https://github.com/rancher/eks-operator/issues/106

Issue #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -310,7 +310,7 @@ func (h *Handler) checkAndUpdate(ctx context.Context, config *eksv1.EKSClusterCo
 		return h.eksCC.UpdateStatus(config)
 	}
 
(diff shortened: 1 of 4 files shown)


salasberryfin (approved):
Thanks @mjura. LGTM. I see this triggers a Go API diff failure, can we keep the same `CheckEBSAddon` and just pass values accordingly?

mjura:
> Thanks @mjura. LGTM. I see this triggers a Go API diff failure, can we keep the same `CheckEBSAddon` and just pass values accordingly?

It will be failing anyway because I have changed header for BuildUpstreamSpec which is used in Rancher. Better is to fix this two problems at once. Anyway it is needed.

yiannistri (approved):

PR #764, opened 2024-08-22
[v2.9] Fixing non-constant format string in call

mjura:
(cherry picked from commit 0ec28e48ed8823444c076da896f550c4665e32c8)

Issue #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -359,7 +359,7 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
 			"must be equal to or one minor version lower than the cluster kubernetes version", aws.ToString(config.Spec.KubernetesVersion), aws.ToString(ng.Version)))
 	}
 	if len(errs) != 0 {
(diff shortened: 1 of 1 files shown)


controller/eks-cluster-config-handler.go
```diff
@@ -359,7 +359,7 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
 			"must be equal to or one minor version lower than the cluster kubernetes version", aws.ToString(config.Spec.KubernetesVersion), aws.ToString(ng.Version)))
 	}
 	if len(errs) != 0 {
-		return fmt.Errorf(strings.Join(errs, ";"))
+		return fmt.Errorf("%s", strings.Join(errs, ";"))
```
yiannistri, line 362: If formatting is not required, perhaps we can use this instead, wdyt? This would require us to also import "errors" package from standard library.
```suggestion
		return errors.New(strings.Join(errs, ";"))
```

yiannistri (approved):

PR #819, opened 2024-09-10
Add debug flag for log level

vatsalparekh:
Add cli flag and set log level debug option. This is helpful when debugging the operator locally. The codebase has debug log statement at multiple places but does not set the log level.


yiannistri (approved):
Thanks! Let's backport this to 2.9 and 2.8 branches too.

PR #833, opened 2024-09-16
update version skew policy

justdan96:
This commit updates the version skew policy to match this document: https://kubernetes.io/releases/version-skew-policy/

Issue #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -353,8 +353,8 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
 		if clusterVersion.EQ(*version) {
 			continue
 		}
-		if clusterVersion.Minor-version.Minor == 1 {
-			continue
+		if (clusterVersion.Minor < 25 && clusterVersion.Minor - version.Minor <= 2) || (clusterVersion.Minor >= 25 && clusterVersion.Minor - version.Minor <= 3) {
+		    	continue
 		}
 		errs = append(errs, fmt.Sprintf("versions for cluster [%s] and nodegroup [%s] not compatible: all nodegroup kubernetes versions"+
 			"must be equal to or one minor version lower than the cluster kubernetes version", aws.ToString(config.Spec.KubernetesVersion), aws.ToString(ng.Version)))


furkatgofurov7:
@mjura @yiannistri folks, can you please take a look when you have time? Thanks

yiannistri:
@justdan96 thank you for the PR, could you also add tests that cover this change? I'll review the change itself in the meantime.

justdan96:
@yiannistri test added

yiannistri:
@mjura I've tested this change locally and it looks fine from my point of view. Can you please also review?

controller/eks-cluster-config-handler.go
```diff
@@ -353,11 +353,16 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
...
 		}
-		if clusterVersion.Minor-version.Minor == 1 {
+		if (clusterVersion.Minor < 25 && clusterVersion.Minor-version.Minor <= 2) || (clusterVersion.Minor >= 25 && clusterVersion.Minor-version.Minor <= 3) {
 			continue
 		}
-		errs = append(errs, fmt.Sprintf("versions for cluster [%s] and nodegroup [%s] not compatible: all nodegroup kubernetes versions"+
-			"must be equal to or one minor version lower than the cluster kubernetes version", aws.ToString(config.Spec.KubernetesVersion), aws.ToString(ng.Version)))
+		if clusterVersion.Minor < 25 {
+			errs = append(errs, fmt.Sprintf("versions for cluster [%s] and node group [%s] are not compatible: for clusters running Kubernetes < 1.25 the "+
+				"node group version may only be up to two minor versions older than the cluster version", aws.ToString(config.Spec.KubernetesVersion), aws.ToString(ng.Version)))
```
mjura, line 361: When this policy was changed, official documentation is still saying that we can't skip any minor version during upgrade? https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/

mjura (changes requested):

controller/eks-cluster-config-handler.go
justdan96: The control plane versions have to be upgraded without skipping minor versions - but this check is comparing the control plane version against the node's kubelet version.

controller/eks-cluster-config-handler.go
```diff
@@ -353,11 +353,16 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
 		if clusterVersion.EQ(*version) {
 			continue
 		}
-		if clusterVersion.Minor-version.Minor == 1 {
+		if (clusterVersion.Minor < 25 && clusterVersion.Minor-version.Minor <= 2) || (clusterVersion.Minor >= 25 && clusterVersion.Minor-version.Minor <= 3) {
```
mjura, line 356: Why it has to be lower than 25? We already dropped support for Kubernetes lower than 27. It doesn't make sense to me

controller/eks-cluster-config-handler.go
```diff
@@ -353,11 +353,16 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
...
 			continue
 		}
-		errs = append(errs, fmt.Sprintf("versions for cluster [%s] and nodegroup [%s] not compatible: all nodegroup kubernetes versions"+
-			"must be equal to or one minor version lower than the cluster kubernetes version", aws.ToString(config.Spec.KubernetesVersion), aws.ToString(ng.Version)))
+		if clusterVersion.Minor < 25 {
```
mjura, line 359: Again this same question, why `25` ?

controller/eks-cluster-config-handler.go
```diff
@@ -353,11 +353,16 @@ func validateUpdate(config *eksv1.EKSClusterConfig) error {
...
+			errs = append(errs, fmt.Sprintf("versions for cluster [%s] and node group [%s] are not compatible: for clusters running Kubernetes < 1.25 the "+
```
mjura, line 360: We already dropped support for Kubernetes lower than 1.27

mjura (changes requested):
We don't want to differentiate between version => v1.25 and lower < v1.25

justdan96:
I can't find a list of supported versions in this repo, are you talking about Rancher itself?

mjura:
> I can't find a list of supported versions in this repo, are you talking about Rancher itself? Even if that's the case the v.2.8.x line still supports K8s 1.25.

Here is the list Kubernetes supported versions in Rancher:
- Rancher v2.9 - https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-9-2/
- Rancher v2.8 - https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-8-8/

This change can be backported and shipped only with next Rancher v2.9 patch version

justdan96:
Sorry I'm still not understanding. If the PR is merged as-is it would be usable with Rancher 2.8.x and Rancher 2.9.x as well. What is the issue?

mjura:
> Sorry I'm still not understanding. If the PR is merged as-is it would be usable with Rancher 2.8.x and Rancher 2.9.x as well. What is the issue?

We don't to hardcore any values in the code of Kubernetes version which are not supported any more by Rancher v2.9 and v2.8. It just doesn't make sense.

justdan96:
Got it, I've now made those changes.

mjura (approved):
Thank you for this change, LGTM

Issue #924, opened 2024-11-06
Trivy timeouts in CI


yiannistri:
The following error appears frequently in our CI logs:
```
ERROR	[vulndb] Failed to download artifact	repo="ghcr.io/aquasecurity/trivy-db:2" err="OCI repository error: 1 error occurred:\n\t* GET https://ghcr.io/v2/aquasecurity/trivy-db/manifests/2: TOOMANYREQUESTS: retry-after: 488.668µs, allowed: 44000/minute\n\n"
FATAL	Fatal error	init error: DB error: failed to download vulnerability DB: OCI artifact error: failed to download vulnerability DB: failed to download artifact from any source
```

PR #925, opened 2024-11-06
ci: Fix trivy action timeout

yiannistri:

We frequently get CI failures when scanning with trivy because there have been too many requests to download the vulnerability db from the rancher org (because many of its repos are using it).

This PR updates the Scan workflow to include additional repositories to be used if trivy fails to retrieve the vulnerability db from the primary source.

Issue #924 

Source for fix: https://github.com/aquasecurity/trivy/discussions/7668#discussioncomment-10884984

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed


mjura (approved):
LGTM

mjura (approved):

Issue #1009, opened 2024-12-11
Suspicious messages in eks-operator logs when importing EKS cluster


thehejik:
On attempt to import an eks cluster into rancher:v2.8-head I see following error messages in eks-operator pod logs.

Versions:
`rancher:v2.8-74d6c7a4c1f564eb2389508e48466590f12fab5a-head`
`rancher-eks-operator:103.6.0+up1.3.6-rc.1`

Command used for provisioning EKS cluster (using latest eksctl 0.198.0):
```bash
eksctl create cluster --name=th-imported --region=eu-west-2 --nodes=1 --node-type=t3.xlarge --node-volume-size=20 --managed --kubeconfig=kubeconfig-th-eks-imported --version 1.28
```

 It is probably harmless and the cluster has been imported successfully, I could scale and add nodepools but the errors seems to be a bit suspicious.
```log
time="2024-12-11T13:36:03Z" level=info msg="Starting eks.cattle.io/v1, Kind=EKSClusterConfig controller"
time="2024-12-11T13:36:03Z" level=info msg="Starting /v1, Kind=Secret controller"
time="2024-12-11T13:36:03Z" level=error msg="Error removing metadata from failure message: message body not formatted as expected"
2024-12-11T13:36:03.955416239Z time="2024-12-11T13:36:03Z" level=error msg="Error recording ekscc [] failure message: resource name may not be empty"
2024-12-11T13:36:04.768479213Z time="2024-12-11T13:36:04Z" level=info msg="cluster [c-5kzvt] finished updating"
2024-12-11T13:36:04.771995869Z time="2024-12-11T13:36:04Z" level=error msg="Error removing metadata from failure message: message body not formatted as expected"
2024-12-11T13:36:04.772034729Z time="2024-12-11T13:36:04Z" level=error msg="Error recording ekscc [] failure message: resource name may not be empty"
2024-12-11T13:36:05.227893452Z time="2024-12-11T13:36:05Z" level=info msg="cluster [c-5kzvt] finished updating"
2024-12-11T13:36:05.755315041Z time="2024-12-11T13:36:05Z" level=info msg="cluster [c-5kzvt] finished updating"
time="2024-12-11T13:38:36Z" level=info msg="waiting for cluster [c-5kzvt] to update nodegroups [th-my-group]"
time="2024-12-11T13:38:37Z" level=info msg="waiting for cluster [c-5kzvt] to update nodegroups [th-my-group]"
[...]

PR #1131, opened 2025-01-30
[main] handle conflict errors while updating eksClusterConfig

vardhaman22:
in case a conflict error occurs the controller's recordError Function will just print it and return the error as it is.

conflict errors occur since rancher also modifies eksconfig object. As conflict error returns empty object, errors like `"Error recording ekscc [] failure message: resource name may not be empty"` occurs when we try to add a failure message to eksconfig status and update it.
The operation which results in conflict gets applied automatically in the next reconciliation call of the controller hence it is safe to just log it. 

Issue # https://github.com/rancher/eks-operator/issues/1009

**Checklist**:

- [X] squashed commits into logical changes
- [ ] includes documentation
- [X] adds unit tests
- [X] backport needed

### DIFF
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -115,7 +115,13 @@ func (h *Handler) recordError(onChange func(key string, config *eksv1.EKSCluster
 			return config, err
 		}
 		if err != nil {
-			if !strings.Contains(err.Error(), "currently has update") {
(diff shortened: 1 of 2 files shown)


mjura (dismissed):
it looks good to me, thank you for adding test coverage

mjura (approved):
LGTM

Issue #752, opened 2024-08-14
EndpointAccessUpdate getting reverted back sporadically


cpinjani:
**Rancher version:**
```
Rancher - v2.9-92f15949d71efb11baf63c878cc2f64bcd25b1e8-head
eks-operator - v1.9.1-rc.6
```
Cluster Type: Downstream EKS cluster

**Describe the bug:**
EndpointAccessUpdate & LoggingUpdate getting reverted back sporadically, probably due to race condition.

For example:
c756d5c6-eca7-3d88-99e8-3e3087078822 - Initial update
5a3cc622-d06d-3fa8-890b-3920ca1c8064 - Reverted back

![image](https://github.com/user-attachments/assets/70a119c9-cff5-45c4-b164-41ccab68f7dd)

**Steps**
- Provision EKS cluster with Public Access - true, Private Access - false
- Edit Private access and set to true

**Logs:**
```
time="2024-08-14T15:07:35Z" level=info msg="cluster [c-m82t7] finished updating"
time="2024-08-14T15:09:16Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:09:16Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:09:46Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:10:17Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:10:47Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:11:17Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:11:47Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:12:18Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:12:52Z" level=info msg="cluster [c-m82t7] finished updating"
time="2024-08-14T15:12:52Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:12:53Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:13:23Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
time="2024-08-14T15:13:53Z" level=info msg="waiting for cluster [c-m82t7] to finish updating"
[...]

PR #1166, opened 2025-02-13
[main] check upstream cluster updates when updating the cluster

vardhaman22:
the cluster refresh controller in rancher reverts some updates to the eks cluster because upstream eks cluster remains in active state for a few seconds after we push an update. 
Added a check to retrieve eks cluster updates and if in case any update is in progress the controller will just return now.

Issue https://github.com/rancher/eks-operator/issues/752

**Checklist**:

- [X] squashed commits into logical changes
- [ ] includes documentation
- [X] adds unit tests
- [X] backport needed

### DIFF
--- a/charts/eks-operator-crd/templates/crds.yaml
+++ b/charts/eks-operator-crd/templates/crds.yaml
@@ -179,6 +179,12 @@ spec:
             type: object
           status:
             properties:
+              completedUpdateIDs:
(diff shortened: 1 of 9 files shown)


mjura (approved):

krunalhinguu (approved):
looks good to me

PR #2131, opened 2026-03-11
fix(eks): correct OIDC provider URL for IPv6 EBS CSI driver role trust policy

iYashMaurya:

For IPv6 EKS clusters with the EBS CSI Driver enabled, the CSI controller pods
were stuck in CrashLoopBackOff with `InvalidIdentityToken` errors.

Root cause: `transformOIDC()` was being applied to the OIDC provider registration
URL, causing a mismatch between:
- The OIDC provider registered in IAM: `oidc-eks.<region>.api.aws`
- The token issuer EKS embeds in service account tokens: `oidc.eks.<region>.amazonaws.com`

AWS STS compares the token issuer against registered OIDC providers and trust
policies — since they didn't match, `AssumeRoleWithWebIdentity` always failed
with `InvalidIdentityToken`.

The fix separates the two concerns:
- `transformOIDC()` is now used **only** for fetching the TLS thumbprint, where
  the IPv6 dual-stack endpoint is needed to get the correct certificate.
- OIDC provider registration and IAM trust policy both use the **original issuer
  URL** from `DescribeCluster`, which matches what EKS embeds in tokens.

Issue https://github.com/rancher/rancher/issues/54029

The key insight is that EKS always issues service account tokens with the
original `oidc.eks.<region>.amazonaws.com` issuer URL regardless of IP family.
The dual-stack `oidc-eks.<region>.api.aws` endpoint exists only for network
connectivity purposes on IPv6 clusters — it should never be used for IAM
provider registration or trust policies.

Before this fix:
```
IAM OIDC Provider:  oidc-eks.<region>.api.aws        ❌
IAM Trust Policy:   oidc-eks.<region>.api.aws        ❌
Token Issuer:       oidc.eks.<region>.amazonaws.com
→ MISMATCH → InvalidIdentityToken
```

After this fix:
```
IAM OIDC Provider:  oidc.eks.<region>.amazonaws.com  ✅
IAM Trust Policy:   oidc.eks.<region>.amazonaws.com  ✅
Token Issuer:       oidc.eks.<region>.amazonaws.com
→ MATCH → EBS CSI Driver Running
```

IPv4 clusters are unaffected since `transformOIDC` was never called for them.

**Checklist**:
- [ ] includes documentation
- [x] adds unit tests
- [ ] backport needed

### DIFF
--- a/pkg/eks/get.go
+++ b/pkg/eks/get.go
@@ -109,5 +109,10 @@ func CheckEBSAddon(ctx context.Context, clusterName string, eksService services.
 		return "", nil
 	}
 
+	if output.Addon.Status == ekstypes.AddonStatusCreateFailed ||
+		output.Addon.Status == ekstypes.AddonStatusDegraded {
+		return "", nil
+	}
+
 	return *output.Addon.AddonArn, nil
 }
--- a/controller/eks-cluster-config-handler.go
+++ b/controller/eks-cluster-config-handler.go
@@ -747,7 +747,7 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, upstreamSpec *
 
 	if templates.IsIPv6(config.Spec.IPFamily) {
 		logrus.Infof("Ensuring OIDC Provider exists for IPv6 cluster [%s]", config.Spec.DisplayName)
-		_, err := awsservices.ConfigureOIDCProvider(ctx, awsSVCs.iam, awsSVCs.eks, config)
+		_, _, err := awsservices.ConfigureOIDCProvider(ctx, awsSVCs.iam, awsSVCs.eks, config)
 		if err != nil {
 			return config, fmt.Errorf("error configuring OIDC provider for IPv6: %w", err)
 		}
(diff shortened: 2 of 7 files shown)


vardhaman22 (approved):

pkg/eks/get.go
```diff
@@ -109,5 +109,10 @@ func CheckEBSAddon(ctx context.Context, clusterName string, eksService services.
 		return "", nil
 	}
 
+	if output.Addon.Status == ekstypes.AddonStatusCreateFailed ||
+		output.Addon.Status == ekstypes.AddonStatusDegraded {
+		return "", nil
+	}
```
yiannistri, line 115: Just curious, why do we check for AddonStatusDegraded here? I thought the addon never gets created in the first place (because of the bug) so wondering if it ever moves to this status.

iYashMaurya: Hey, 
The addon actually does get created at the API level (Deployments, etc., are applied). However, because of the OIDC bug, the underlying pods fail STS auth and get stuck in CrashLoopBackOff. When EKS sees the pods crashing continuously, it transitions the overall addon status to DEGRADED rather than CREATE_FAILED. That's exactly why we need to catch this specific state.

yiannistri (approved):

